PR #29, opened 2011-06-28
Add the method default_option in form generate by crud

nicodmf:
I have idea to include others default inclusion :
- collection for many to many and one to many
- subform for one to one

Do i make?


javiereguiluz:
Although this pull request is still open, the changes proposed in it are already merged in the bundle's code. So we can close it as "merged elsewhere". Thanks.

PR #32, opened 2011-06-30
Changing generated route /hello/{name} to /

gregquat:
The goal is to have a generated code more "neutral"and functional without change.
It's just a proposal, I hope I did not forget anything modifying it.

Grégory Quatannens


javiereguiluz:
@gregquat thanks for starting this pull request. I've rebased and tweaked your code in this other PR: #383. That's why I'm closing this without merging it. I've taken your ocmmits, so you'll get the full credit for your work. Thanks.

Issue #67, opened 2011-08-25
Proposal: Remove DependencyInjection dir


weaverryan:
Hey guys!

I'd like to propose that the bundle generator does *not* generate a `DependencyInjection` directory. My thoughts are that creating a DI extension class is an advanced thing, and that most end-users should simply be importing DI resource config files directly from `config.yml`. Furthermore, I think - like routing config - that we should teach that just one config file is loaded by symfony and that you must import all other resources from that file. Having a DI extension that automatically loads a `services.xml` clouds this.

Thanks!

javiereguiluz:
This issue can be closed because #312 was already merged.

PR #89, opened 2012-01-03
[GeneratorBundleCommand] refactored code to enable customization via inheritance.

lazaridis-com:
Bug fix: no
Feature addition: yes
Backwards compatibility break: no
Symfony2 tests pass: should (not tested)
Fixes the following tickets: #67 (provides workaround)
Todo: possibly add example file (exists already)

Issues like #67 can be solved easily by customization via inheritance on a language level.

In order to allow this, GenerateBundleCommand gets some protected variables.

This is a refactoring, which does not alter the existent behaviour.


javiereguiluz:
Is this feature request still relevant or can we think of better customization/overriding mechanisms?

javiereguiluz:
Closign it in favor of the other pending PRs, specially #235.

PR #93, opened 2012-01-08
added a generator:bundle:thin command in order to generate minimal bundles

lazaridis-com:
**Bug fix:** no
**Feature addition:** yes
**Backwards compatibility break:** no
**Symfony2 tests pass:** should (not tested)
**Fixes the following tickets:** #67
**Todo:** add a test (I'll provide one, if code get's accepted)

Adds a "generator:bundle:thin", using the existent source codes and skeletons.

The benefit is that users can choose to generate a "thin" bundle, which contains only the most basic files.

Language level inheritance is used, leaving all other source files untouched.


javiereguiluz:
This issue can be closed because the bundle generator already allows to not generate the whole dir structure (and the latest changes allow to generate very thin bundles: without vendor name, without `DependencyInjection/` dir, etc.)

PR #99, opened 2012-01-24
[Skeleton] Check if datetime fields are defined as the same way of index.html.twig skeleton

Maks3w:
show.html.twig doesn't check if datetime is defined before to use `date` filter.

Added the same check than already implemented in index.html.twig


javiereguiluz:
I've rebased and finished this PR in #388. @Maks3w I always take the original author's commit to give him/her the full credit of your work. In this case I did the same ... but since your change was made in a now-inexistent template (`show.html.twig` was replaced by `show.html.twig.twig`) your commit has "disappeared". I'm sorry about that.

PR #102, opened 2012-01-25
Unified generator interface/implementation

tanakahisateru:
* Form generator behavior was changed to interactive.
* Form generator argument(entity) was changed to optional.
* Entity generator and form generator were refactored to be similar to others.

## before:

```
app/console doctrine:generate:form AcmeBundle:Post
```

## after:

```
app/console doctrine:generate:form --entity=AcmeBundle:Post
```

or

```
app/console doctrine:generate:form
```

then specify target entity name interactively.

### DIFF
--- a/Command/GenerateDoctrineEntityCommand.php
+++ b/Command/GenerateDoctrineEntityCommand.php
@@ -31,13 +31,13 @@ class GenerateDoctrineEntityCommand extends GenerateDoctrineCommand
     protected function configure()
     {
         $this
-            ->setName('doctrine:generate:entity')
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Closing it because it's better if some elements which are mandatory remain as arguments instead of options. As explained by @stof, we can have arguments + options and still be interactive for all those values.

stof:
@tanakahisateru interaction to ask for missing arguments should be done in the ``interact`` method which is called before validating the input (it is called only in interactive mode)

PR #109, opened 2012-02-05
introduced automatically adding a twig extends tag and blockname

skoop:
One of the things I've noticed when generated basic CRUD pages for applications I've already been working on for a while is the first thing I do: add extends and block tags to all generated templates. 

This pull request removes the need to do that every time you've used the CRUD generation, by enabling an optional template name and block name. By specifying the template name and block name, the templates that are generated will extend the specified base template and all page output will be placed between a block tag.


javiereguiluz:
I'm closing this issue because there are a lot of other issues and pull requests that will allow to customize these options and that do it in a more generic way. Thanks.

PR #156, opened 2012-10-05
Fixed duplicate route generation on regeneration with existing routes

lighthart:
Bug Fix: Yes- previous versions make a duplicate of the route if doctrine:generate:entities is called to update an existing entity.


javiereguiluz:
Closing it as "fixed" because the reported problem was already fixed in the pull request #381.

PR #201, opened 2013-06-14
add a remove bundle

inalgnu:


javiereguiluz:
I propose to close this issue because the proposed method is not strong enough (it relies on `str_replace()`). Besides, in https://github.com/symfony/symfony/issues/6082 there is a more general discussion about install/uninstall bundles, so it's better to wait for the final decision. Thanks.

PR #203, opened 2013-07-01
[Skeleton][Bundle] Add valid html via extends base to default template

ErichHartmann:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no ticket found
| License       | MIT
| Doc PR        | n/a

Currently, the skeleton for the Bundle Generator produces an index.html.twig file that does not include standard HTML tags. Symfony2 profiler/toolbar requires standard tags to display the need elements. In other words, the default controller and view created by the Bundle Generator will not show the profiler/toolbar when making a request via app_dev.php.

By adding the extends '::base.html.twig' and the proper title and body twig elements, the default controller and template produced by the Bundle Generator displays the proper profiler/toolbar elements when using app_dev.php. The result is a valid html page all other things being equal.


javiereguiluz:
The problem with this proposal is that you cannot assume that the `base.html.twig` exists in the application or whether it's stored at that location (`::base.html.twig`) or whether it defines those blocks (`title` and `body`). So loading this page could lead to a Twig error.

Therefore, even if the original idea of this pull request is right, I propose it to close it as "won't fix".

PR #210, opened 2013-07-11
Adding a --no-controller flag to generate:bundle command

seiffert:
Passing the `--no-controller` flag to the generate:bundle command prevents
that the DefaultController class with an example action is being created. If
no controller class is being created, the routing config file will be omitted
as well and the app's routing config won't be updated.

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -40,6 +40,7 @@ class GenerateBundleCommand extends GeneratorCommand
                 new InputOption('bundle-name', '', InputOption::VALUE_REQUIRED, 'The optional bundle name'),
                 new InputOption('format', '', InputOption::VALUE_REQUIRED, 'Use the format for configuration files (php, xml, yml, or annotation)'),
                 new InputOption('structure', '', InputOption::VALUE_NONE, 'Whether to generate the whole directory structure'),
+                new InputOption('without-controller', '', InputOption::VALUE_NONE, 'Whether to generate an example controller with a first route.'),
             ))
             ->setDescription('Generates a bundle')
             ->setHelp(<<<EOT
@@ -101,6 +102,7 @@ EOT
         }
         $format = Validators::validateFormat($input->getOption('format'));
         $structure = $input->getOption('structure');
+        $controller = !$input->getOption('without-controller');
 
         $dialog->writeSection($output, 'Bundle generation');
 
@@ -109,7 +111,7 @@ EOT
         }
 
         $generator = $this->getGenerator();
-        $generator->generate($namespace, $bundle, $dir, $format, $structure);
+        $generator->generate($namespace, $bundle, $dir, $format, $structure, $controller);
 
         $output->writeln('Generating the bundle code: <info>OK</info>');
 
@@ -122,8 +124,10 @@ EOT
         // register the bundle in the Kernel class
         $runner($this->updateKernel($dialog, $input, $output, $this->getContainer()->get('kernel'), $namespace, $bundle));
 
-        // routing
-        $runner($this->updateRouting($dialog, $input, $output, $bundle, $format));
+        if ($controller) {
+            // routing
+            $runner($this->updateRouting($dialog, $input, $output, $bundle, $format));
+        }
 
         $dialog->writeGeneratorSummary($output, $errors);
     }
@@ -239,6 +243,14 @@ EOT
             $structure = true;
         }
         $input->setOption('structure', $structure);
+
+        $withoutController = $input->getOption('without-controller');
+        if (!$withoutController && $dialog->askConfirmation($output, $dialog->getQuestion('Do you want to generate an example controller with a first route?', 'yes', '?'), false)) {
+            $withoutController = false;
+        } else {
+            $withoutController = true;
+        }
+        $input->setOption('without-controller', $withoutController);
 
         // summary
         $output->writeln(array(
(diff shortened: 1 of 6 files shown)


Command/GenerateBundleCommand.php
```diff
@@ -101,6 +102,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
         $format = Validators::validateFormat($input->getOption('format'));
         $structure = $input->getOption('structure');
+        $withController = !$input->getOption('no-controller');
```
OskarStark: i would change ```$withoutController``` to ```$noController```

Command/GenerateBundleCommand.php
```diff
@@ -240,6 +244,12 @@ protected function interact(InputInterface $input, OutputInterface $output)
         }
         $input->setOption('structure', $structure);
 
+        $withController = !$input->getOption('no-controller');
```
OskarStark: why do you define the variable again?

Generator/BundleGenerator.php
```diff
@@ -28,7 +28,7 @@ public function __construct(Filesystem $filesystem)
         $this->filesystem = $filesystem;
     }
 
-    public function generate($namespace, $bundle, $dir, $format, $structure)
+    public function generate($namespace, $bundle, $dir, $format, $structure, $controller)
```
OskarStark: i would rename ```$controller``` to ```$renderController```

what about using ```true``` as default?

Generator/BundleGenerator.php
```diff
@@ -46,19 +46,23 @@ public function generate($namespace, $bundle, $dir, $format, $structure)
...
-            'bundle'    => $bundle,
-            'format'    => $format,
-            'bundle_basename' => $basename,
-            'extension_alias' => Container::underscore($basename),
+            'namespace'                     => $namespace,
+            'bundle'                        => $bundle,
+            'format'                        => $format,
+            'bundle_basename'               => $basename,
+            'extension_alias'               => Container::underscore($basename),
+            'include_example_controller'    => $controller,
```
OskarStark: ``ìnclude_example_controller``` to ```render_controller``` ?

Generator/BundleGenerator.php
```diff
@@ -66,7 +70,7 @@ public function generate($namespace, $bundle, $dir, $format, $structure)
             $this->renderFile('bundle/services.'.$format.'.twig', $dir.'/Resources/config/services.'.$format, $parameters);
         }
 
-        if ('annotation' != $format) {
+        if ('annotation' != $format && $controller) {
```
OskarStark: i would check the $controller var first in the if condition

Tests/Generator/BundleGeneratorTest.php
```diff
@@ -17,7 +17,7 @@ class BundleGeneratorTest extends GeneratorTest
 {
     public function testGenerateYaml()
     {
-        $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false);
+        $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false, true);
```
OskarStark: you can keep the previous version if you define it default to ```true```, otherwise its a breaking change for users who implement the generate()-method somewhere else

Tests/Generator/BundleGeneratorTest.php
```diff
@@ -55,13 +80,22 @@ public function testGenerateAnnotation()
...
+    }
+
     public function testDirIsFile()
     {
         $this->filesystem->mkdir($this->tmpDir.'/Foo');
         $this->filesystem->touch($this->tmpDir.'/Foo/BarBundle');
 
         try {
-            $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false);
+            $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false, true);
```
OskarStark: same as above

Tests/Generator/BundleGeneratorTest.php
```diff
@@ -74,7 +108,7 @@ public function testIsNotWritableDir()
         $this->filesystem->chmod($this->tmpDir.'/Foo/BarBundle', 0444);
 
         try {
-            $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false);
+            $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false, true);
```
OskarStark: same as above

Tests/Generator/BundleGeneratorTest.php
```diff
@@ -88,7 +122,7 @@ public function testIsNotEmptyDir()
         $this->filesystem->touch($this->tmpDir.'/Foo/BarBundle/somefile');
 
         try {
-            $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false);
+            $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false, true);
```
OskarStark: same as above

Tests/Generator/BundleGeneratorTest.php
```diff
@@ -100,7 +134,7 @@ public function testExistingEmptyDirIsFine()
     {
         $this->filesystem->mkdir($this->tmpDir.'/Foo/BarBundle');
 
-        $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false);
+        $this->getGenerator()->generate('Foo\BarBundle', 'FooBarBundle', $this->tmpDir, 'yml', false, true);
```
OskarStark: same as above

OskarStark:
good work, @fabpot what do you think about my comments?

javiereguiluz:
@OskarStark in my opinion this should be closed as "won't fix". Reasons:

  * If we provide a `--no-controller` option, should we also provide `--no-*` options for the rest of the bundle elements?
  * In my experience, 99% of application bundles include one or more controllers. My guess is that this would be a "niche option".

OskarStark:
:+1:  @javiereguiluz

Issue #213, opened 2013-07-20
Generating entities with spaces in field names creates syntax errors


OskarStark:
This:
![bildschirmfoto 2013-07-20 um 09 37 36](https://f.cloud.github.com/assets/995707/829895/9fb7dac4-f110-11e2-998b-d6a81239280f.png)

creates this:
![bildschirmfoto 2013-07-20 um 09 36 45](https://f.cloud.github.com/assets/995707/829896/a9365184-f110-11e2-86c3-4374eac6bde9.png)

javiereguiluz:
Closing it as "fixed" by #394.

PR #214, opened 2013-07-20
Field names with spaces causing syntax errors. 

OskarStark:
See #213


javiereguiluz:
I've been digging a bit about this. The database column names are less restrictive than entity property names. The reason is that properties are PHP variables and column names are database responsibility.

For example, you can use `call-limit` as a column name, but not as a property name (see http://www.doctrine-project.org/jira/browse/DDC-1288).

So we could apply the usual regular expression to check if some value is a valid PHP variable name:

```php
private function isValidFieldName($name)
{
    return preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $name, $matches);
}
```

@OskarStark, what do you think about giving a final boost to this two-year old issue to finish it? Thanks!

OskarStark:
i absolutely agree with you @javiereguiluz 

my problem is that the current code (as i said) can produce syntax errors.
IMO we should be failsafe on the php level, the database itself should throw an error if the fieldname is invalid.

but the goal for me is that the syntax should be correct anyway

OskarStark:
@javiereguiluz shall i provide a PR with this change?

from where do you get this method?

javiereguiluz:
Yes please, provide a PR and I'll review it. I took the method from StackOverflow and verified it with the PHP manual.

OskarStark:
@javiereguiluz i added you as collaborator.

i commited my changed in the branch ```fieldname-check````

when i add phpunit the test will stuck for a long time...
when i execute the tests against ```master``` it is very fast.

can you please have a look?
maybe there are some problems withe the closure and the referenced ```$matches``` var in the ```preg_match```

thank you

OskarStark:
bildschirmfoto 2015-07-19 um 17 43 57
bildschirmfoto 2015-07-19 um 17 46 31

OskarStark:
hmm, with this commit the tests are running @javiereguiluz 

https://github.com/OskarStark/SensioGeneratorBundle/commit/79c6a430a0f992dd4440144f87b93010b3ad9300

OskarStark:
i added a new PR with the idea of @javiereguiluz here: #394

javiereguiluz:
Closing it in favor of #394

PR #215, opened 2013-07-30
Remove typo in twig body block ('-')

idciconsulting:
This is just a detail !


javiereguiluz:
Technically it's not a typo, but I agree with @idciconsulting: this `-` should be removed because the generated templates don't include any other white space operator (we should fix everything or nothing, but not just one occurrence).

javiereguiluz:
Thanks @idciconsulting.

PR #216, opened 2013-07-30
[DX] [CRUD] Added support for entity name autocompletion in the CRUD generator

csarrazi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

As in the doctrine:generate:entity command, there is now support for entity name autocompletion in the form ```VendornameBundlenameBundle:MyEntity```.
Entities are filtered and fetched from the src/ folder and thus, vendor entities are excluded. Repositories are also excluded from the entity list. Autocompletion works well, even entities which are in sub-directories.

### DIFF
--- a/Command/AutoComplete/EntitiesAutoCompleter.php
+++ b/Command/AutoComplete/EntitiesAutoCompleter.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\GeneratorBundle\Command\AutoComplete;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * Provides auto-completion suggestions for entities.
+ *
+ * @author Charles Sarrazin <charles@sarraz.in>
+ */
+class EntitiesAutoCompleter
+{
+    private $manager;
+
+    public function __construct(EntityManagerInterface $manager)
+    {
+        $this->manager = $manager;
+    }
+
+    public function getSuggestions()
+    {
+        $configuration = $this->manager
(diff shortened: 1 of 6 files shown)


csarrazi:
Pong @weaverryan

csarrazi:
Ping anyone? ;)

weaverryan:
Hey @csarrazi - thanks for pinging me - I'll take a look at this soon - my schedule has been crushing me lately - but this is in my inbox/on my mind!

csarrazi:
No harm done, @weaverryan !
I do know what it is to have a very busy schedule! ;)

weaverryan:
Hey @csarrazi!

This looks really really nice, and it's working *great* for me. Thanks for your patience on this with me :). The only change I personally want is to remove the `sensio_generator.command.auto_complete.entities` service, and have the command just instantiate this directly (and so, you might also remove `getAutocompleteEntities` from `GenerateDoctrineCommand`, otherwise you'd be instantiating an object there for 1 use, which is odd). I have two reasons: (A) I don't want to make the container bigger for a service that's almost never used (and only used in dev) and (B) *technically* if a user has multiple em's, this service only works for the "default" em. I don't think we should worry about supporting a non-default em in this command (or if we do, do it later), but if we keep this service, other people might start using it, and then eventually someone might want to create one of these services for *each* em. I think it's better to keep it just as a class, and other people can register it as a service if they want to use it.

If you agree and can make this change, then I think we'll be ready!

Thanks!

csarrazi:
I'm fine with the change. I'll push an update later during the week! :-)

csarrazi:
Done, @weaverryan!

weaverryan:
:+1: from me, and I just tested it - works great! And fast updates too @csarrazi :)

This is ready for a merge.

fabpot:
Thank you @csarrazi.

PR #240, opened 2013-11-12
make the force overwrite flag also overwrite generated forms when generating crud

edmondscommerce:
Modification so that using the force overwrite flag will make the generator overwrite form classes as well as the controller

Without the flag an exception will be thrown and displayed to the user (rather than silently caught) if a form class already exists. They can then decide to run with force overwrite or take any other action


javiereguiluz:
Closing it for being finished in #400.

PR #246, opened 2013-12-03
Adding Autocomplete of Entity shortcut names to doctrine:generate:crud Command

thaberkern:
Adding Autocomplete of Entity shortcut names to doctrine:generate:crud Command


javiereguiluz:
Closing it because entity name autocompletion is already enabled in the command. Thanks.

Issue #248, opened 2013-12-08
Change form generator function from private to protected to allow override and extend from the sub class


jamesmoey:
I want to be able to extend or override function in form generator so that I can implement custom function.

PR #249, opened 2013-12-10
Change form generator function from private to protected to allow override and extend from the sub class

jamesmoey:
Issue #248

### DIFF
--- a/Generator/DoctrineFormGenerator.php
+++ b/Generator/DoctrineFormGenerator.php
@@ -92,7 +92,7 @@ class DoctrineFormGenerator extends Generator
      * @param  ClassMetadataInfo $metadata
      * @return array             $fields
      */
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Before accepting/rejecting this proposal, I'd like to ask the opinion of @stof, which is our biggest expert in deciding if some feature should be public/private/protected. Thanks!

stof:
what is your use case for making this an extension point ? Your issue just says ``I want this to become an extension point``, but extension points involve BC concerns, which is why we always want to think about the right way to provide the extension point (most of the time, inheritance is not the answer to this question). And we cannot think about it without knowing the use case

xabbuh:
I suggest to close here. I doubt @jamesmoey still needs this after almost two years.

PR #251, opened 2013-12-19
Kernel manipulator improvements:

alsma:
Hello! 
Adding new bundle using SensioGeneratorBundle, could provoke syntax error in case when $bundles array does not contain any bundle (e.g. they added dynamically in parent kernel class). So this pull request improves code generation and cover found issues and most relevant cases by unit tests.

    - change generation algorithm
    - Added test to cover KernelManipulator
    - Added stub objects to cover most relevant cases


javiereguiluz:
@alsma I'm very sorry for the late response to your pull request.

Your work was almost done, but it needed some rebase and minor tweaks. I've done that in #406 ... but maintaining all your original commits so you get full credit for your work. Thanks!

Issue #244, opened 2013-11-28
[Console] generate:bundle does not work with XML config


cordoval:
"Auto-update of routing does not work, when using routing.xml. It generates new routing.yml file and writes into it. If no auto-update of routing is issued generate:bundle returns Yaml.
routing.yml file seems to be hardcoded into the module with no plans of supporting XML config in the future (even the documentation of Symfony says that users can use Xml as well and are not limited to Yaml).

Steps to reproduce the bug:
1. Create routing.xml with routing schema.
2. Change config.xml or config.yml or config.php to include rounting.xml in routing:config
3. Use generate:bundle command.

Symfony version: 2.0.12
PHP Version: 5.3.10
Apache Version: 2.2.22
OS version: 3.0.0-17-generic (Ubuntu 11.10)

Bug fix: yes
Todo: PHP config auto-update. I use only XML config, so I fix it, but you should support and php config as well."

copied from https://github.com/symfony/symfony/issues/3796

/cc @IlluminatiBG

JHGitty:
Why is this issue not be done yet? It's open since 2 years.

firestorm23:
Confirming the bug. Just started using Symfony2 and then this bug appears. Not good, especially when I expected good tools from this framework to not doing repeating work and duplicating code.

JHGitty:
@firestorm23: The Symfony Best Practices are using YAML for the routing file(s). That because this issue is not widely known I think. But yes: It should be fixed!

firestorm23:
@JHGitty I know. Maybe it's just me, but something tells me, that this feature is too long and boring to fix. Yaml is very simple and convenient, but XML put you in constraints and schema restrictions. I myself prefer an XML, cause it's more solid, but it requires more code writing to parse and build dom. So i think this is one of those feature which wasn't done in time and no one wants to pick it up.

PR #252, opened 2013-12-20
#244 [Console] generate:bundle does not work with XML config

podolskis:
I've changed generating routing part - no yml data printing to xml, php or .annotation files. No routing file for annotations. Please have a look at it, I would appreciate feedback.

Detailed problem description is  here - https://github.com/sensiolabs/SensioGeneratorBundle/issues/244


javiereguiluz:
Closing it because of the lack of feedback and the concerns raised by reviewers.

PR #254, opened 2013-12-24
It would be nice, if you could merge my appending files in the new and edit view rendering.

3web:


javiereguiluz:
Closing it as "fixed". The field mapping is already passed to most views and in #401 we pass it to the "new" view, the only missing view.

PR #256, opened 2014-01-15
Creating views with proper arrays and booleans 

piotrpasich:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

I'd like to catch two cases - when you've got an array there should be shown all values separeted by coma in the index and view template. Really similar situation is with boolean - I wouldn't like to see 0 or 1 value because I'm always going to change it to yes or no. So, I've added this case too.


javiereguiluz:
I think this change makes a lot of sense. Since no one has commented against this, I'm going to merge it.

javiereguiluz:
Thanks @piotrpasich.

PR #262, opened 2014-02-17
Update crud generation with fields definitions WIP

nicodmf:
For many sites, it is impossible to use the form twig method only.

Defined each fields with custom html markup is necessary.

Without this change it is impossible to just overwrite template, and the only solution is to recreate another command, so add fields to all generation function permit to simply use twig template inheritence without hard coding many codes.

This change was human tested, it work.


javiereguiluz:
Closing it as "fixed". The field mapping is already passed to most views and in #401 we pass it to the "new" view, the only missing view.

PR #288, opened 2014-07-05
Added support for entity name autocompletion in the CRUD generator. Alternative to https://github.com/sensiolabs/SensioGeneratorBundle/pull/216

borisson:
Not sure if the strpos / substr method i've used is the way to go but it works


javiereguiluz:
Closing it because entity name autocompletion is already enabled in the command. Thanks.

PR #289, opened 2014-07-05
[DX] Potential missing annotation route import

benji07:
Register controller route when a bundle is generated in xml, yml and php, but we want to generate a crud with annotation

### DIFF
--- a/Manipulator/RoutingManipulator.php
+++ b/Manipulator/RoutingManipulator.php
@@ -12,6 +12,7 @@
 namespace Sensio\Bundle\GeneratorBundle\Manipulator;
 
 use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\Yaml\Yaml;
 
 /**
  * Changes the PHP code of a YAML routing file.
@@ -88,11 +89,17 @@ class RoutingManipulator extends Manipulator
             return false;
         }
 
-        $current = file_get_contents($this->file);
+        $config = Yaml::parse($this->file);
 
-        $search = 'resource: "@%s/Controller/"';
+        $search = sprintf('@%s/Controller/', $bundle);
 
-        return false !== strpos($current, $search);
+        foreach ($config as $resource) {
+            if (array_key_exists('resource', $resource)) {
+                return $resource['resource'] == $search;
+            }
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Manipulator/RoutingManipulator.php
```diff
@@ -74,4 +75,63 @@ public function addResource($bundle, $format, $prefix = '/', $path = 'routing')
...
+        }
+
+        $code = sprintf("%s:\n", Container::underscore(substr($bundle, 0, -6)).'_'.Container::underscore($controller));
+
+        $code .= sprintf("    resource: \"@%s/Controller/%sController.php\"\n    type:     annotation\n", $bundle, $controller);
+
+        $code .= "\n";
+        $code .= $current;
+
+        if (false === file_put_contents($this->file, $code)) {
```
phansys: Why not just ```return false !==  file_put_contents($this->file, $code);```?

Manipulator/RoutingManipulator.php
```diff
@@ -74,4 +75,63 @@ public function addResource($bundle, $format, $prefix = '/', $path = 'routing')
...
+            return false;
+        }
+
+        $config = Yaml::parse($this->file);
+
+        $search = sprintf('@%s/Controller/', $bundle);
+
+        foreach ($config as $resource) {
+            if (array_key_exists('resource', $resource)) {
+                return $resource['resource'] == $search;
```
phansys: Strict comparison should be used.

Manipulator/RoutingManipulator.php
```diff
@@ -74,4 +75,63 @@ public function addResource($bundle, $format, $prefix = '/', $path = 'routing')
...
+     *
+     * @return bool
+     */
+    public function hasResourceInAnnotation($bundle)
+    {
+        if (!file_exists($this->file)) {
+            return false;
+        }
+
+        $config = Yaml::parse($this->file);
```
phansys: >Passing a filename is deprecated in Symfony 2.2, and will be removed in Symfony 3.0.
[[Ref]](http://symfony.com/doc/current/components/yaml/introduction.html).

javiereguiluz:
@benji07 thanks for your great work on this. Since only a final push was needed to finish it, I've decided to create a new PR in #403. I reused all your commits, so you'll get full credit on your work. Thanks!

benji07:
@javiereguiluz :+1:

PR #296, opened 2014-07-13
Services configuration format choice for bundle generation with the 'annotation' format

Djuuu:
When you chose 'annotation' as the generated configuration files format, 'xml' is used for the services configuration file and there is no way to chose another format.

With this patch, when 'annotation' is chosen, the user will be asked to chose a format for the services configuration.

I'm not quite sure about the etiquette with pull-requests here (maybe I should open an issue first ?), and this patch might not be really elegant, but here is my proposition anyway :-)

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -266,11 +266,15 @@ EOT
         $input->setOption('structure', $structure);
 
         // summary
+        $services = '';
+        if ($servicesFormat != $format) {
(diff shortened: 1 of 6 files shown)


javiereguiluz:
Thanks for proposing this change. I'm closing this pull request without merging it because another pull request already did this change recently. Instead of adding a new configuration format, I think it's better to comply with the official best practices and use `services.yml` when using the annotations format. By the way, the current code has a minor bug relate to this and I'm trying to fix it in #384.

Issue #293, opened 2014-07-07
[DX] Display the list of available entities


javiereguiluz:
The command asks you to write the shortcut name of the entity before generating its CRUD:

```cli
$  php app/console doctrine:generate:crud

  Welcome to the Doctrine2 CRUD generator

This command helps you generate CRUD controllers and templates.

First, you need to give the entity for which you want to generate a CRUD.
You can give an entity that does not exist yet and the wizard will help
you defining it.

You must use the shortcut notation like AcmeBlogBundle:Post.

The Entity shortcut name: AcmeDemoBundle:Entity
```

This could be improved if the command displayed the list of available entities:

```cli
$  php app/console doctrine:generate:crud

Welcome to the Doctrine2 CRUD generator

Select the entity for which you want to generate CRUD
controllers and templates:

[1] AcmeBlogBundle:Post
[2] AcmeBlogBundle:Comment
[3] AcmeBlogBundle:Tags
[4] AcmeBlogBundle:User
[5] AcmeAdminBundle:Entity
[6] AcmeAdminBundle:Invoices
[7] ...

Type the name of the entity: 

(if the entity doesn't exist yet, bootstrap it with the
doctrine:generate:entity command)
```

javiereguiluz:
We recently decided to refocus the purpose of this bundle to not complicate it too much. Given that the current command already autocompletes the bundle + entity name, we think that it's not necessary to display the full list of application entities to pick one. That's why I'm closing this and all the other related issues and pull requests as "won't fix". Thanks for understanding it.

WouterJ:
@javiereguiluz this PR doesn't widen the scope of the bundle. It only makes the current feature more easy to use.

Autocompletion doesn't work on Windows, so this is very useful for Windows devs.

javiereguiluz:
My reasoning:

1) This is a nice but optional feature. Everything works perfectly without it.
2) This feature works like a charm on Linux and Mac.

So my questions are:

1) Is this an unavoidable limitation of Windows systems?
2) Is this a bug/limitation of Symfony Console component?

WouterJ:
If it's a nice feature, I see no point in closing the issue... (especially not closing as won't fix)

Afaik, there is no way to make autocompletion work on Windows. However, I have an idea for an autocomplete-ish feature for the Console component which works on Windows, I'll submit a PR to Symfony about it somewhere next week.

PR #300, opened 2014-08-24
[DX] Improving generate:doctrine:crud #293

adridev:
Improve "generate:doctrine:crud" command adding two features:
    - Add option "--with-list", which shown a list of all available entities with an index. User should select one and then proceed with generation of the CRUD.
    - Add auto-completion when user inserts entity name. 

| Q                       | A
| -------------          | ---
| Bug fix?             | no
| New feature?    | yes
| BC breaks?       | no
| Deprecations?  | no
| Tests pass?      | yes
| Fixed tickets      |
| License             | MIT
| Doc PR             |

### DIFF
--- a/Command/GenerateDoctrineCrudCommand.php
+++ b/Command/GenerateDoctrineCrudCommand.php
@@ -235,6 +235,8 @@ EOT
 
     protected function getEntitiesList()
     {
+
+        /** @var \Doctrine\ORM\EntityManager $entityManager */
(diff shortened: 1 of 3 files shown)


javiereguiluz:
We recently decided to refocus the purpose of this bundle to not complicate it too much. Given that the current command already autocompletes the bundle + entity name, we think that it's not necessary to display the full list of application entities to pick one. That's why I'm closing this and all the other related issues and pull requests as "won't fix". Thanks for understanding it.

Issue #297, opened 2014-08-19
Impossible to generate CRUD for entities inherited


maidmaid:
If I have an entity class Cheese that inherits of entity Products, the CRUD Command does not take into account fields of Products. Metadata used doesn't load the informations of super class. The metadata factory used is ```DisconnectedMetadataFactory``` (in ```GenerateDoctrineCommand::getEntityMetadata()```). If I replace it with ```\Doctrine\Bundle\DoctrineBundle\Mapping\MetadataFactory```, it works.

javiereguiluz:
I'm closing this issue as "won't fix" because we are rethinking the purpose of this bundle to not complicate it too much. Entity inheritance is one of the edge features that we've decided to not support for now. Thanks for understanding it.

PR #302, opened 2014-09-17
[console] Fix bundle selection in controller generation

eschmar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

When generating a new controller for a bundle which has derivations (another bundle inherits from this bundle), the function ``getBundle()`` executed from kernel will return the child. Consequently the controller will be created in the wrong bundle.

[The kernel comments](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Kernel.php#L464) state, that the order of the bundleMap hierarchy is ``most derived bundle first``. Without the ``false`` flag, the function ``getBundle`` will only return the first element. The PR simply chooses the last element.

I'm not sure if [#297](https://github.com/sensiolabs/SensioGeneratorBundle/issues/297) is relating to this problem, however i've seen the ``getBundle`` command in the ``\Command\Helper\GenerateDoctrineCrudCommand.php`` class as well. Unfortunately I don't have the time right now to check this.

//edit: I'm a little confused why the unit test fails.. help would be appreciated!

### DIFF
--- a/Command/GenerateControllerCommand.php
+++ b/Command/GenerateControllerCommand.php
@@ -88,7 +88,8 @@ EOT
             $bundle = Validators::validateBundleName($bundle);
 
             try {
-                $bundle = $this->getContainer()->get('kernel')->getBundle($bundle);
+                $bundleMap = $this->getContainer()->get('kernel')->getBundle($bundle, false);
+                $bundle = end($bundleMap);
             } catch (\Exception $e) {
                 $output->writeln(sprintf('<bg=red>Bundle "%s" does not exist.</>', $bundle));
             }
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm closing this issue as "won't fix" because we are rethinking the purpose of this bundle to not complicate it too much. Bundle inheritance is one of the edge features that we've decided to not support for now. Thanks for understanding it.

PR #312, opened 2014-11-19
Update generate:bundle for non-shared bundles and more

weaverryan:
Hi guys!

If you're reading this, please try this out and offer any feedback! There are so many different cases, it will make @fabpot's life much much easier (and that's always good).

This replaces (adds to) @rafix's work on #299. The task ended up being a bit larger, because there are so many things we *could* change, and we need to communicate things well.

Fortunately, the changes here alter the generation process very little - most changes are to *how* we ask the questions. Here are the highlights:

* if a bundle is non-shared, we only ask for a bundle name, not a namespace
* removed "generate full directory structure" functionality entirely. This added only a few extra files, which I don't think add value.
* make `.yml` the service configuration format if you choose annotation format. If you're using annotation for routing, it's likely you're following the best practices are you might be surprised by XML :)
* many small output changes

BC breaks:

* `BundleGenerator::generate` is now gone
* The `structure` option was removed
* You now get `services.yml` if you choose annotation as your format

In #299, we asked the question "should a non-shared bundle have a `Resources/` directory?". I think it should. Why? I think the `AppBundle` is special - it's the one bundle that's truly coupled to your kernel (e.g. `app/` directory), which is why it's ok to store the templates in `app/`. If we also stored templates from other bundles in `app/`, it gets a crowded. If you had a `DefaultController` in 2 bundles, their templates would "want" to occupy the same space (e.g. `app/Resources/views/default/*`).  So, I think it makes sense to have only one special bundle (per kernel) that you really *couple* to the kernel. *IF* you choose to add other bundles (which you don't need to do), I think you should organize them more along the lines of traditional bundle, at least for templates.

### TODOS

* [x] Import services.xml/yml manually from config.yml for non-shared bundle
* [x] Can we break BC in the ways described above? Or do I need to re-add somethings?
* [x] Fix the GenerateBundleCommandTest
* [x] Add more cases to BundleGeneratorTest for no DI directory
* [x] Fix BundleGeneratorTest

What does everyone think? /cc @javiereguiluz 

Thanks!

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -11,6 +11,7 @@
 
 namespace Sensio\Bundle\GeneratorBundle\Command;
 
+use Sensio\Bundle\GeneratorBundle\Manipulator\ConfigurationManipulator;
 use Sensio\Bundle\GeneratorBundle\Model\Bundle;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
@@ -102,7 +103,12 @@ EOT
         // routing importing
         $runner($this->updateRouting($output, $bundle));
 
-        $questionHelper->writeGeneratorSummary($output, $errors);
+        if (!$bundle->shouldGenerateDependencyInjectionDirectory()) {
+            // we need to import their services.yml manually!
+            $runner($this->updateConfiguration($output, $bundle));
+        }
+
+        $dialog->writeGeneratorSummary($output, $errors);
     }
 
     protected function interact(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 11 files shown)


WouterJ:
ping @weaverryan

fabpot:
@weaverryan Can you rebase this PR?

weaverryan:
This is ready to go again! And I have tested all the different paths I can find. Highlights:

1) Added the `shared` option

2) Removed the `structure` option

3) The task now ALWAYS asks questions (even if the values are passed as args) unless you explicitly say `--no-interactive` (this drastically simplified the code)

4) For non-shared bundles, you don't get a `DependencyInjection` directory, and your `services.yml` file is imported directly from `config.yml`.

5) The signature to `BundleGenerator::generateBundle()` changed - there is now a nice `Bundle` class instead of passing around all the config for a bundle.

@javiereguiluz if you have a few minutes to try this, it would probably help Fabien have the confidence to merge (since the diff is too big to really study).

Thanks!

fabpot:
As there are some "BC breaks", I've bumped the version to 3.0: 176ea172c9df471fd71eb9e6c0e5992f3cc06626

fabpot:
... which makes merging easier :)

fabpot:
Thanks @weaverryan for working on this feature, this is much appreciated.

fabpot:
It's in now, but if @javiereguiluz would test it thoroughly, that would be tremendously helpful (we can even tweak things as we are working on a version nobody is going to use before we decide to tag 3.0.0).

PR #322, opened 2014-12-08
fix warning message coloring for existing form

smatyas:
This PR fixes a warning message in the doctrine:generate:crud command, as the `<warning>` console formatter does not exist. (Maybe it should exist in the symfony/Console, as I'd like to use it a lot myself as well, but that's an other question :smile: )

I assumed to use the `<comment>` formatter for warnings, but we can create a custom `<warning>` formatter if you want.

The output looks like this before the fix:
```
Generating the Form code: <warning>Already exists, skipping</warning>
```

Additionally, a help message typo is also fixed.


fabpot:
Thank you @smatyas.

PR #328, opened 2015-01-11
Kernel bundle manipulation compatible with short array syntax

thanosp:
Bundle generation right now fails if you switch AppKernel to short array syntax.

- Made checks compatible with both short and traditional syntax
- Corrected type in docblock to for CS test to pass


fabpot:
Thank you @thanosp.

PR #330, opened 2015-01-13
remove useless return value

henter:


fabpot:
This is a template and the `$config` variable is almost always needed. Not in this example, but we must keep it to make it clear that the configuration is in this variable and you should get it from there.

henter:
get it, thank you @fabpot :+1:

PR #334, opened 2015-02-09
Remove usage of deprecated _method requirement in favor of methods option

WouterJ:


WouterJ:
@fabpot fabbot should be updated to not consider `.yml.twig` as being valid YAML files (Twig syntax isn't valid Yaml).

fabpot:
Thank you @WouterJ.

PR #340, opened 2015-03-31
Make generated classes final by default.

WouterSioen:
Resources:

http://ocramius.github.io/blog/when-to-declare-classes-final/
http://verraes.net/2014/05/final-classes-in-php/

### DIFF
--- a/Resources/skeleton/bundle/Bundle.php.twig
+++ b/Resources/skeleton/bundle/Bundle.php.twig
@@ -7,7 +7,7 @@ use Symfony\Component\HttpKernel\Bundle\Bundle;
 {% endblock use_statements %}
 
 {% block class_definition %}
-class {{ bundle }} extends Bundle
+final class {{ bundle }} extends Bundle
 {% endblock class_definition %}
 {
 {% block class_body %}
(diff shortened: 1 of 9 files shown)


WouterSioen:
@fabpot @weaverryan any opinions on this?

weaverryan:
Hmm I'm -1, mostly because I don't see much benefit - final is more important when you're building code you'll share. I think a lot of users will wonder what the significance is of the final- why you need it in symfony (You don't need it, but when people see it, they might assume so).

For me, little benefit, but it may confuse people. People who understand final can add it if they wish to.

Cheers!

WouterSioen:
Hmm, I get the confusing users part, but it's a part of the language, so I guess PHP developers should not be confused by it, and if they are, it's time for them to learn the keyword.

The final keyword is ment to show that a class should not be extended. Controllers are one of the best examples of classes you'll never extend. If there's one type of classes in your application that are logical to have the final keyword, it's your controller classes.

But again, I get your point too, and I get it if you close this PR. It would be just nice to see the final keyword a little more in PHP code. It's great to enforce composition instead off massive inheritance.

fabpot:
Closing this PR as I share @weaverryan's point of view.

PR #342, opened 2015-04-06
Generate entity repositories in 'Repository' namespace

xelaris:
This helps to keep the *Entity* namespace free from non-entity classes and to use the same structure as in the new Symfony demo application (see https://github.com/symfony/symfony-demo/issues/24). It should be BC, since the Repositories are still generated in *Entity* namespace per default.

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -22,7 +22,6 @@ use Symfony\Component\HttpKernel\KernelInterface;
 use Sensio\Bundle\GeneratorBundle\Generator\BundleGenerator;
 use Sensio\Bundle\GeneratorBundle\Manipulator\KernelManipulator;
 use Sensio\Bundle\GeneratorBundle\Manipulator\RoutingManipulator;
-use Sensio\Bundle\GeneratorBundle\Command\Helper\QuestionHelper;
 
 /**
  * Generates bundles.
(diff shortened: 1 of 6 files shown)


weaverryan:
@xelaris I'd say, remove the new question, and *always* put the class in the `Repository` namespace. I don't think BC is too important with a generator, and more importantly, I recently had some BC-breaking changes merged into the master branch, and we're planning on releasing a new version soon (so, old versions are BC, new version will break slightly).

I want to keep things simple :)

fabpot:
I agree with @weaverryan

xelaris:
I like that. The PR is updated.

gcds:
I came to write the same issue then saw this... Yes and please remove that question from generator or just make it default yes because who do not use Repository? :D

weaverryan:
I agree with @gcds! Especially since the repositories will be in their own directory, they're not "taking up space". So, I would:

A) Automatically generate the repositories (don't ask the question)

AAAND, if you're up for it @xelaris, I think it would also be cool to:

B) We should print out a better summary on the bottom, that describes that the Entity file was created, (if YML or XML format) the *.xml/yml file was created and the Repository was created.

Currently, the command finishes like this:
screen shot 2015-05-03 at 4 49 41 pm

I'd like to print out the above information - basically make it look like the end of the `generate:bundle` command that's on the master branch:

screen shot 2015-05-03 at 4 48 06 pm

@xelaris What do you think about these?

xelaris:
@weaverryan Sounds good to me. The PR is updated to address both points. In addition I removed the "Summery before generation" and its confirmation, like it was done for the bundle generation command - hoping it's welcome.
bildschirmfoto 2015-05-04 um 01 21 34

fabpot:
Thank you @xelaris.

weaverryan:
Wow, great work (and super fast) @xelaris!

PR #345, opened 2015-05-04
Update generate_doctrine_entity.rst

andesk:
There is no routing in doctrine entities, instead the format is for entity mapping configuration. Seems to have its source from documentation of generate:bundle command.


fabpot:
Thank you @andesk.

PR #292, opened 2014-07-06
[DX] generate:crud minor improvements

musonic:
This relates to [issue 283](https://github.com/sensiolabs/SensioGeneratorBundle/issues/283). Rename $entity to something real - like $product or $blogPost for a BlogPost entity

PR #346, opened 2015-05-05
Fix template vars of crud views

xelaris:
As of #292 the variables in the generated templates don't match the variables passed in the generated controller.

```
Variable "entities" does not exist in src/AppBundle/Resources/views/Subject/index.html.twig
```


fabpot:
Thank you @xelaris.

Issue #344, opened 2015-04-27
The crud-generator doesn't follow symfony best practices


emil-nasso:
Are there any plans to change the generator so that it follows the best practices available at http://symfony.com/doc/current/best_practices/index.html ?

To my understanding a new demo application will be released with symfony 2.7.  ( http://symfony.com/blog/introducing-the-symfony-demo-application)

The demo application follows the best practices recommendations, it would be really nice if the generator would too.

As a know symfony developer i started out by reading the book and then the best practices document and experimented a bit with it. I felt like i had a good idea on how to write symfony applications. Then i used the crud-generator and had a "what the hell"-moment. ;)

JHGitty:
+1 :+1: 
I was going to create the same issue today and found this issue.

```
Don't use the @Template() annotation to configure the template used by the controller.
```
http://symfony.com/doc/master/best_practices/controllers.html#template-configuration

emil-nasso:
I think the largest issue is that it uses three methods for each action. For creating an entity, there is `createCreateForm`, `createAction` and `newAction`. The best practices does this with one method in a much less verbose and clean way. This takes a bit of work to refactor so i usually just write everything myself, from scratch.

JHGitty:
@emil-nasso same here!

Lewik:
@emil-nasso +1

maidmaid:
Here are my suggestions for improvements:

1) [So it's much simpler to let newAction handle everything.](http://symfony.com/doc/current/best_practices/forms.html#handling-form-submits)

While with current generator :

```php
public function createAction(Request $request) {}
public function newAction() {}
```

2) [Add buttons in the templates, not in the form classes or the controllers.](http://symfony.com/doc/current/best_practices/forms.html#form-button-configuration)

While with current generator :

```php
$form->add('submit', 'submit', array('label' => 'Create'));
```

3) [Store all your application's templates in app/Resources/views/ directory](http://symfony.com/doc/current/best_practices/templates.html#template-locations)

While with current generator, templates is stored in AppBundle/Resources/views/

4) [Rendering the Form](http://symfony.com/doc/current/best_practices/forms.html#rendering-the-form)

```twig
{{ form_start(form, {'attr': {'class': 'my-form-class'} }) }}
    {{ form_widget(form) }}
{{ form_end(form) }}
```

While with current generator :

```twig
{{ form(edit_form) }}
```

[...]

PR #347, opened 2015-05-06
Remove usage of @Template to follow best practice

xelaris:
relates to https://github.com/sensiolabs/SensioGeneratorBundle/issues/344#issuecomment-98558657


fabpot:
Thank you @xelaris.

PR #349, opened 2015-05-14
Extend Routing Manipulator when $prefix is given

frankadrian:
When using the $prefix parameter generated routing identifiers can sometimes be malformed.
for example providing: `/{_locale}/project/` as the prefix, the route name would result in: `..._{_locale}_project_`
This fixes the trailing `_` by first checking the prefix for a slash.
Additionally it removes the {_locale} parameter if it exists.


javiereguiluz:
@frankadrian thanks for starting this fix. I'm finishing it in #393. I've taken all your commits, so you'll get full credit for your work when it is merged. Thanks!

Issue #10, opened 2011-06-23
doctrine:generate:entities and OneToMany


abdeltiflouardi:
Hello all,

There is no way to generate OneToMany by command?

javiereguiluz:
Recently we decided to refocus the purpose of this bundle and work only on its core features. Supporting entity associations is one of the features that we prefer to not implement yet. The alternative solution is to use third-party bundles such as SonataAdmin, which support this and many other complex features.

Therefore, I'm closing this and all the other related issues and pull requests as "won't fix". Thanks for understanding it.

Issue #278, opened 2014-06-18
Improving doctrine:generate:entity add mapping association


recchia:
## The Problem

The command to generate entities is very useful but when you need to map associations have to do it by hand, which is very boring and tedious

## The Solution

Add another data type called entity requesting the information for generating the association as the name of the entity with its name space and the type of relationship as a minimum. But it could also be added depending on the type of relationship more specific attributes.

For example:

type field: entity
type of relation: manytoone

optional:

mappingby
inversedby
joincolumn
jointable

as appropriate

PD: Please put DX label

zeopix:
https://github.com/sensiolabs/SensioGeneratorBundle/pull/352

javiereguiluz:
Recently we decided to refocus the purpose of this bundle and work only on its core features. Supporting entity associations is one of the features that we prefer to not implement yet. The alternative solution is to use third-party bundles such as SonataAdmin, which support this and many other complex features.

Therefore, I'm closing this and all the other related issues and pull requests as "won't fix". Thanks for understanding it.

PR #352, opened 2015-05-30
Added relations support to entity mapping generator

zeopix:


Command/AutoComplete/EntitiesAutoCompleter.php
```diff
@@ -27,25 +31,34 @@ public function __construct(EntityManagerInterface $manager)
...
-
         $entities = $configuration
             ->getMetadataDriverImpl()
             ->getAllClassNames()
         ;
 
-        return array_map(function ($entity) use ($namespaceReplacements) {
-            return strtr($entity, $namespaceReplacements);
-        }, $entities);
+        if ($applyNamespaceReplacements == true) {
```
phansys: Please, use Yoda condition and strict comparison.

Command/GenerateDoctrineEntityCommand.php
```diff
@@ -216,9 +218,28 @@ private function addFields(InputInterface $input, OutputInterface $output, Quest
         }
         $output->writeln('');
 
-        $fieldValidator = function ($type) use ($types) {
+        $output->write('<info>Available relations:</info> ');
+
+        $relations = array('many_to_many', 'many_to_one', 'one_to_one', 'one_to_many');
+        foreach ($relations as $i => $relation) {
+            $count += strlen($relation);
```
phansys: What is the indended use for this? You're counting how many chars have the relationship's names and this var seems to be unused.

Command/GenerateDoctrineEntityCommand.php
```diff
@@ -216,9 +218,28 @@ private function addFields(InputInterface $input, OutputInterface $output, Quest
...
+            $output->write(sprintf('<comment>%s</comment>', $relation));
+            if (count($relations) != $i + 1) {
```
phansys: Shouldn't be easier something like ```$output->write('<comment>'.implode('</comment>, <comment>', $relations).'</comment>.';```

Command/GenerateDoctrineEntityCommand.php
```diff
@@ -289,6 +310,88 @@ private function addFields(InputInterface $input, OutputInterface $output, Quest
...
             }
 
+            if ($type == 'one_to_many') {
+                $question = new Question($questionHelper->getQuestion('Target Entity class', ''), '');
+                $question->setAutocompleterValues($autocompleteEntities);
+                $data['targetEntity'] = $questionHelper->ask($input, $output, $question);
+
+                $question = new Question($questionHelper->getQuestion('Mapped by', ''), '');
+                $mappedBy = $questionHelper->ask($input, $output, $question);
+                if (empty($mappedBy) == false)
```
phansys: Please, use braces.

Command/GenerateDoctrineEntityCommand.php
```diff
@@ -289,6 +310,88 @@ private function addFields(InputInterface $input, OutputInterface $output, Quest
                 $data['length'] = $questionHelper->ask($input, $output, $question);
             }
 
+            if ($type == 'one_to_many') {
+                $question = new Question($questionHelper->getQuestion('Target Entity class', ''), '');
+                $question->setAutocompleterValues($autocompleteEntities);
```
phansys: The user input should be validated.

Command/GenerateDoctrineEntityCommand.php
```diff
@@ -289,6 +310,88 @@ private function addFields(InputInterface $input, OutputInterface $output, Quest
...
+                    if (empty($inversedBy) == false)
+                        $data['inversedBy'] = $inversedBy;
+
+                    $question = new Question($questionHelper->getQuestion('Join Table Name', ''), '');
+                    $joinTableName = $questionHelper->ask($input, $output, $question);
+                    if (empty($joinTableName) == false)
+                        $data['joinTable'] = array('name' => $joinTableName);
+
+                    $data['isOwningSide'] = true;
+                }else{
```
phansys: Please, respect [PSR-2 CS](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#51-if-elseif-else).

Command/GenerateDoctrineEntityCommand.php
phansys: You should do just a strict comparison of the user's input against the default value.

phansys:
These options should be supported in ```--no-interactive``` mode and some tests are worth.

phansys:
Closes #10, #278.

javiereguiluz:
Recently we decided to refocus the purpose of this bundle and work only on its core features. Supporting entity associations is one of the features that we prefer to not implement yet. The alternative solution is to use third-party bundles such as SonataAdmin, which support this and many other complex features.

Therefore, I'm closing this and all the other related issues and pull requests as "won't fix". Thanks for understanding it.

ryall:
@javiereguiluz Is there a reason you don't consider entity associations a core feature when they are used in a high percentage of all entity classes?

PR #357, opened 2015-06-17
[GenerateDoctrineEntityCommand] Added support for more field definitions

### LINKED ISSUE #343: [Command] generate nullables, and precision/scale for decimals
Everytime I generate an entity, I have to manually go through every field and set the precision/scale (in the case of decimal type) and set the <code>nullable=true</code> for the fields that can accept NULL. Is it possible that the <code>php app/console doctrine:generate:entity</code> prompts for all these properties? Or, since the nullable is almost always false, have an argument like <code>--prompt-nullables</code> so it prompts only on that entity?

phansys:
Added support for field definitions like ```nullable```, ```unique```,  ```precision```, ```scale```

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #343 
| License       | MIT
| Doc PR        |

Closes #343.

Allows to define if a field is nullable and precision/scale properties for decimals:
```
New field name (press <return> to stop adding fields): some_field
Field type [string]:
Field length [255]:
Is nullable [false]:
Unique [false]:

New field name (press <return> to stop adding fields): other_field
Field type [string]: decimal
Precision [2]:
Scale [1]:
Is nullable [false]:
Unique [false]:
```

To keep consistency with current results, it only writes ```nullable=true``` or  ```unique=true``` in schema definition when the choosed option is ```true```. If they have ```false``` as value, the command will completely omit these definitions.

**TODO:**
- [x] Add tests
- [x] Add support for ```unique``` definition
- [x] Add support for passing definition within ```--fields``` argument


phansys:
I need thoughts about the format in which this definition will be received through command arguments, since I found a little hard to find a good way given the current syntax:
```
php app/console doctrine:generate:entity --entity=AcmeBlogBundle:Blog/Post --format=annotation --fields="title:string(255) body:text" --with-repository --no-interaction
```

phansys:
I think it could be resolved with the following (verbose) syntax:
```bash
--fields="title:string(length=255 nullable=true) rank:decimal(precission=10 scale=0 nullable=true)"
```
while keeping BC compatibility with current short syntax for ```string``` type when there is no need to supply these options:
```bash
--fields="title:string(255) rank:decimal(precission=10 scale=0 nullable=true)"
```

Any thoughts are welcome.

phansys:
This PR is ready for review.

fabpot:
Looks good, :+1:

phansys:
Can I squash my commits?

fabpot:
No need to squash as I can do that easily when merging.

phansys:
Excellent!

fabpot:
Thank you @phansys.

PR #348, opened 2015-05-06
Generate configureOptions method on form type

xelaris:
This avoids the generation of deprecated `setDefaultOptions` method in SF 2.7. If the `configureOptions` method isn't available, as in SF 2.6, the generated form type will be forward compatible.

PR #359, opened 2015-06-22
avoid deprecated method

garak:


xabbuh:
`configureOptions()` is not available on Symfony 2.6.

garak:
@xabbuh this bundle is currently requiring ~2.6

phansys:
@xabbuh, the latest release for Symfony standard 2.6 (2.6.9) has a [requirement for ```~2.3```](https://github.com/symfony/symfony-standard/blob/v2.6.9/composer.json#L24) and this PR was made against master.

xabbuh:
That's true, but nothing is preventing one to use the bundle with Symfony 2.3.

phansys:
I think that premise applies to a lot of other situations, but I agree that here could be worst since this is a code generator tool.
Maybe is possible to check what version of Symfony is installed and generate the appropriate code block for each case.

phansys:
I just found a previous PR addressing this.
Duplicate of #348, which is also compatible with `< 2.7`.

garak:
Thanks @phansys

PR #361, opened 2015-06-23
[DoctrineEntityGeneratorTest] Removed 2nd argument for DoctrineEntityGeneratorTest::generate()

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Removed 2nd argument for ```DoctrineEntityGeneratorTest::generate()``` method
since the support for this flag was removed in 80c2cb7dd3ccf2262aa9d1f77f6e8a4103aabac3.

PR #362, opened 2015-06-23
[GenerateDoctrineEntityCommand] Added option to set id generator strategy

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Allows to specify identifier generator strategy while creating a new entity:

```
$ app/console doctrine:generate:entity

  Welcome to the Doctrine2 entity generator

This command helps you generate Doctrine2 entities.

First, you need to give the entity name you want to generate.
You must use the shortcut notation like AcmeBlogBundle:Post.

The Entity shortcut name: AcmeBlogBundle:Post

Determine the format to use for the mapping information.

Configuration format (yml, xml, php, or annotation) [annotation]:

Determine the identifier generation strategy to use as entity id.

Identifier generation strategy (AUTO, SEQUENCE, IDENTITY, UUID, TABLE or NONE) [AUTO]:

Instead of starting with a blank entity, you can add some fields now.
Note that the primary key will be added automatically (named id, which uses the AUTO generation strategy).

Available types: array, simple_array, json_array, object,
boolean, integer, smallint, bigint, string, text, datetime, datetimetz,
date, time, decimal, float, binary, blob, guid, json.

New field name (press <return> to stop adding fields):

```

It should be merged after #361 since it uses the 5th argument for ```DoctrineEntityGenerator::generate()```, which previously was used for ```$with_repository``` param.

### DIFF
--- a/Command/GenerateDoctrineEntityCommand.php
+++ b/Command/GenerateDoctrineEntityCommand.php
@@ -36,6 +36,7 @@ class GenerateDoctrineEntityCommand extends GenerateDoctrineCommand
             ->addOption('entity', null, InputOption::VALUE_REQUIRED, 'The entity class name to initialize (shortcut notation)')
             ->addOption('fields', null, InputOption::VALUE_REQUIRED, 'The fields to create with the new entity')
             ->addOption('format', null, InputOption::VALUE_REQUIRED, 'Use the format for configuration files (php, xml, yml, or annotation)', 'annotation')
+            ->addOption('id-generator-strategy', null, InputOption::VALUE_REQUIRED, 'The identifier generation strategy (AUTO, SEQUENCE, IDENTITY, UUID, TABLE, NONE)', 'AUTO')
             ->setHelp(<<<EOT
 The <info>doctrine:generate:entity</info> task generates a new Doctrine
 entity inside a bundle:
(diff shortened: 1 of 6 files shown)


phansys:
NOTE: fabbot.io fail isn't related to these changes.

phansys:
NOTE: travis fail isn't related to these changes, it was a composer timeout in 5.5 matrix.

javiereguiluz:
As you may know, recently we decided to refocus the purpose of this bundle and work only on its core features. I believe that the proposed `id-generator-strategy` option is too specific and uncommon to fit in those "core features". Therefore, I'm closing it as "won't fix". Thanks for understanding it.

phansys:
Hi @javiereguiluz. I agree that some features are really out of the scope of this bundle, but I think this one isn't since I think it is a little bit arbitrary to disallow the option to choose the desired Id strategy generation (since I'm using UUID).
I read your comment at other PR (#352) and I have no idea which is the equivalent command for this in SonataAdmin: https://sonata-project.org/bundles/admin/master/doc/reference/console.html.

PR #364, opened 2015-06-28
Change form skeleton to new Symfony 2.7 pattern

merorafael:
Change form skeleton to use configureOptions method and OptionsResolver class, default in Symfony 2.7.


phansys:
Duplicate of #359.

PR #369, opened 2015-07-01
Update setDefaultsOptions() method

mzmgh:
As you now setDefaultOptions() method was introduced in Symfony 2.7. Previously, the method was called setDefaultOptions() but when I use doctrine:generate:crud command it create setDefaultOptions() method instead of configureOptions() method.
for use configureOptions() method use Symfony\Component\OptionsResolver\OptionsResolver must be add.


xabbuh:
see #359

PR #374, opened 2015-07-04
[2.4] [Validators] Added basic check for entity name

### LINKED ISSUE #70: Doctrine:Entity Generator does not evaluate correctly either the Entity Shortcut name
Hello,

There is a bug when you build an organization from the console and put the shortcut name without the name of the entity after the colon.

Example:
The Entity shortcut name: AQPHostAQPSearchBundle:

Determine the format to use for the mapping information.

Configuration format (yml, xml, php, or annotation) [annotation]: dominio

Then in the folder src / AQPHost / AQPSearchBundle / Entity is not created entity only the repository with the name "Repository.php"

--------------------------Generator Historial

ratasxy@ratasxy-aqphost:~/Proyectos/aqpsearch$ php app/console doctrine:generate:entity

                                             
  Welcome to the Doctrine2 entity generator  
                                             

This command helps you generate Doctrine2 entities.

First, you need to give the entity name you want to generate.
You must use the shortcut notation like AcmeBlogBundle:Post.

The Entity shortcut name: AQPHostAQPSearchBundle:

Determine the format to use for the mapping information.

Configuration format (yml, xml, php, or annotation) [annotation]: dominio
 Format "dominio" is not supported. 
Configuration format (yml, xml, php, or annotation) [annotation]:     

Instead of starting with a blank entity, you can add some fields now.
Note that the primary key will be added automatically (named id).

Available types: array, object, boolean, integer, smallint, 
bigint, string, text, datetime, datetimetz, date, time, decimal, float.

New field name (press <return> to stop adding fields): dominio
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): cuenta
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): reseller
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): fecha
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): server
Field type [string]:     
Field length [255]: 

New field name (press <return> to stop adding fields): tipo
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): removedor
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): comentario
Field type [string]: 
Field length [255]: 

New field name (press <return> to stop adding fields): date
Field type [string]: date

New field name (press <return> to stop adding fields): 

Do you want to generate an empty repository class [no]? yes

                             
  Summary before generation  
                             

You are going to generate a "AQPHostAQPSearchBundle:" Doctrine2 entity
using the "annotation" format.

Do you confirm generation [yes]? 

                     
  Entity generation  
                     

Generating the entity code: OK

                                               
  You can now start using the generated code!  
                                               
-------------------------------------

Juan Afredo Salas Santillana

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #70

Fixes #70.


Command/Validators.php
```diff
@@ -96,11 +96,23 @@ public static function validateFormat($format)
...
+     * @param string $entity
+     *
+     * @return string
+     *
+     * @throws \InvalidArgumentException
+     */
     public static function validateEntityName($entity)
     {
         if (false === strpos($entity, ':')) {
             throw new \InvalidArgumentException(sprintf('The entity name must contain a : ("%s" given, expecting something like AcmeBlogBundle:Blog/Post)', $entity));
```
fabpot: Can we remove this one?

phansys: Yes, since it's checked by the regex. Done.
It was kept here because this check is so specific, thus I thought that maybe somebody trust this exception message.

fabpot:
I've merged this one on master as we don't maintain older branches.

fabpot:
Thank you @phansys.

phansys:
Thank you @fabpot.

PR #383, opened 2015-07-16
Simplify the initial code of generated bundles

javiereguiluz:
In the past, generated bundles were a good opportunity to show Symfony features (such as the routing parameters). Nowadays we have the pre-generated AppBundle and the Symfony Demo application, so we can make the generated bundles as clean as possible.


fabpot:
Thank you @javiereguiluz.

PR #384, opened 2015-07-17
Generate services.yml instead of services.xml when using annotations

javiereguiluz:
In the past, `services.xml` was generated when using the annotation format. This was changed recently to generate `services.yml` instead. The problem is that the DI extension class was still trying to load `services.yml`.


fabpot:
Thank you @javiereguiluz.

PR #259, opened 2014-01-20
Fixed creating actions from command-line

kucharovic:
<table>
<thead><tr>
<th>Q</th>
<th>A</th>
</tr></thead>
<tbody>
<tr>
<td>Bug fix?</td>
<td>yes</td>
</tr>
<tr>
<td>New feature?</td>
<td>no</td>
</tr>
<tr>
<td>BC breaks?</td>
<td>no</td>
</tr>
<tr>
<td>Deprecations?</td>
<td>no</td>
</tr>
<tr>
<td>Tests pass?</td>
<td>yes</td>
</tr>
<tr>
<td>License</td>
<td>MIT</td>
</tr>
</tbody>
</table>

Fixed generating controller from command-line with option `--actions`

PR #385, opened 2015-07-17
Fixed the generation of controller actions

javiereguiluz:
This PR finishes #259 with some minor tweaks.


fabpot:
Thank you @kucharovic.

PR #386, opened 2015-07-17
generate:crud command now only works for existing entities

### LINKED ISSUE #71: automatic entity creation issue (in CRUD mode)
`php app/console gen:doctrine:crud` says the following:

```
This command helps you generate CRUD controllers and templates.

First, you need to give the entity for which you want to generate a CRUD.
You can give an entity that does not exist yet and the wizard will help
you defining it.
```

I'm issuing the above command with the `AcmeBlogBundle:Post` entity name (`AcmeBlogBundle` does exist, `Post` does not exist, no any other entities in `AcmeBlogBundle`). Generator answers the following

```
[Doctrine\ORM\ORMException]

Unknown Entity namespace alias 'AcmeBlogBundle'.
```

It only allows to proceed with CRUD generation after explicit Post creation (`php app/console gen:doctrine:entity`). If I have defined entities in `AcmeBlogBundle`, error message is different though the issue stays - wizard doesn't let me automatically create entity during crud generation as it claims to:

```
[Doctrine\ORM\Mapping\MappingException]

Class Acme\BlogBundle\Entity\Post is not a valid entity or mapped super class.

Tested on symfony-standard 2.0.0, 2.0.3 and current version from the Symfony repository.
PHP 5.3.5-1ubuntu7.2 with Suhosin-Patch (cli) (built: May 2 2011 23:00:17)
```

javiereguiluz:
This fixes #71.

### DIFF
--- a/Command/GenerateDoctrineCrudCommand.php
+++ b/Command/GenerateDoctrineCrudCommand.php
@@ -112,7 +112,7 @@ EOT
             throw new \RuntimeException(sprintf('Entity "%s" does not exist in the "%s" bundle. Create it with the "doctrine:generate:entity" command and then execute this command again.', $entity, $bundle));
         }
 
-        $bundle      = $this->getContainer()->get('kernel')->getBundle($bundle);
+        $bundle = $this->getContainer()->get('kernel')->getBundle($bundle);
 
         $generator = $this->getGenerator($bundle);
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @javiereguiluz.

PR #388, opened 2015-07-17
Make the "show" template check datetime fields before displaying them

javiereguiluz:
This is consistent with the behavior of the `index` template.


fabpot:
Thank you @javiereguiluz.

PR #393, opened 2015-07-20
Fix route names and route file importing when route prefix is complex

javiereguiluz:
This finishes #349.

-----

### Before

```
Determine the routes prefix (all the routes will be "mounted" under this
prefix: /prefix/, /prefix/new, ...).

Routes prefix [/post]: /{_locale}/project/
```

The generated route names were:

```php
    /**
     * Displays a form to create a new Post entity.
     *
     * @Route("/new", name="{_locale}_project__new")
     * @Method("GET")
     * @Template()
     */
    public function newAction() { ... }
```

In addition, when using YAML, the error message also contained a wrong YAML key:

bad_error

### After

We now sanitize the route prefix to generate good route names:

```php
    /**
     * Displays a form to create a new Post entity.
     *
     * @Route("/new", name="project_new")
     * @Method("GET")
     * @Template()
     */
    public function newAction() { ... }
```

And also good error messages:

good_error


Generator/DoctrineCrudGenerator.php
```diff
@@ -296,4 +293,14 @@ protected function getRecordActions()
             return in_array($item, array('show', 'edit'));
         });
     }
+
+    public static function getRouteNamePrefix($prefix)
+    {
+        $prefix = str_replace(array('{_locale}', '{_format}'), array('', ''), $prefix);
```
fabpot: what about removing anything enclosed in `{}` instead of hardcoding those two?

javiereguiluz: Much better! Done. Thanks.

Generator/DoctrineCrudGenerator.php
```diff
@@ -296,4 +293,14 @@ protected function getRecordActions()
             return in_array($item, array('show', 'edit'));
         });
     }
+
+    public static function getRouteNamePrefix($prefix)
+    {
+        $prefix = preg_replace('/{(.*?)}/', '', $prefix);   // {foo}_bar -> _bar
+        $prefix = str_replace('/', '_', $prefix);
+        $prefix = preg_replace('/(_)\\1+/', '$1', $prefix); // foo__bar -> foo_bar
```
fabpot: should be simplified with `preg_replace('/_+/', '_', $prefix)`... or am I missing something?

javiereguiluz: Much simpler. I like it! Thanks.

PR #395, opened 2015-07-20
Validate entity property names

javiereguiluz:
This refactores #394.

### DIFF
--- a/Command/Validators.php
+++ b/Command/Validators.php
@@ -127,6 +127,19 @@ class Validators
         return $entity;
     }
 
+    public static function validateEntityPropertyName($property)
+    {
+        if (in_array($property, self::getReservedWords())) {
+            throw new \InvalidArgumentException(sprintf('Name "%s" is a PHP reserved word.', $property));
+        }
+
+        if (!preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $property)) {
+            throw new \InvalidArgumentException(sprintf('The property name isn\'t valid (it can only contain letters, numbers and underscores; it must start with a letter)', $property));
+        }
+
+        return $property;
+    }
+
     public static function getReservedWords()
     {
         return array(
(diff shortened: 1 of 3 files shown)


Command/Validators.php
```diff
@@ -127,6 +127,19 @@ public static function validateEntityName($entity)
         return $entity;
     }
 
+    public static function validateEntityPropertyName($property)
+    {
+        if (in_array($property, self::getReservedWords())) {
+            throw new \InvalidArgumentException(sprintf('Name "%s" is a PHP reserved word.', $property));
+        }
+
+        if (!preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $property)) {
```
OskarStark: if you use the pattern directly, you can remove the method from the generator class

Command/Validators.php
```diff
@@ -127,6 +127,19 @@ public static function validateEntityName($entity)
         return $entity;
     }
 
+    public static function validateEntityPropertyName($property)
+    {
+        if (in_array($property, self::getReservedWords())) {
```
OskarStark: if we use this, we shall remove the isReservedKeyword()-method, but this is a BC

OskarStark:
looks like the is_null() check is necessary, because travis is running and running :)

OskarStark:
my PR #394 passed all the test now :)

OskarStark:
@javiereguiluz that is the same problem i had on my local machine

bildschirmfoto 2015-07-22 um 10 36 58

i don't know exactly why, but if you change this:
```php
if (!preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $property)) {
```

to this:
```php
if (!is_null($property) && !preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $property)) {
```

it should be working!

javiereguiluz:
Closing it in favor of #394.

PR #396, opened 2015-07-20
[minor] Conversion to "int" and "bool" for Doctrine generated entities

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/Generator/DoctrineEntityGenerator.php
+++ b/Generator/DoctrineEntityGenerator.php
@@ -88,6 +88,11 @@ class DoctrineEntityGenerator extends Generator
             $entityGenerator->setGenerateAnnotations(false);
             $entityCode = $entityGenerator->generateEntityClass($class);
         }
+        $entityCode = str_replace(
+            array("@var integer\n", "@var boolean\n", "@param integer\n", "@param boolean\n", "@return integer\n", "@return boolean\n"),
+            array("@var int\n", "@var bool\n", "@param int\n", "@param bool\n", "@return int\n", "@return bool\n"),
+            $entityCode
+        );
 
         $this->filesystem->mkdir(dirname($entityPath));
         file_put_contents($entityPath, $entityCode);


javiereguiluz:
Thanks for this commit. I like the idea behind it, but it looks like a hack. Apparently the code we are fixing is the same code that we generated (via Doctrine libraries). It's strange that we generate wrong code in the first place :confused:

phansys:
Thank you for the feedback @javiereguiluz. I've created a PR last week against Doctrine in order to "fix" this into the library itself: doctrine/doctrine2#1457. I'm quoting fix because this CS rule isn't part of Doctrine's CSs:
* https://github.com/doctrine/doctrine2/blob/master/CONTRIBUTING.md#coding-standard
* https://github.com/doctrine/coding-standard/blob/master/Docs/README.md#doctrine-coding-standard

javiereguiluz:
:+1: then it looks like this is the only safe way to solve this issue. Thanks.

Generator/DoctrineEntityGenerator.php
```diff
@@ -88,6 +88,11 @@ public function generate(BundleInterface $bundle, $entity, $format, array $field
             $entityGenerator->setGenerateAnnotations(false);
             $entityCode = $entityGenerator->generateEntityClass($class);
         }
+        $entityCode = str_replace(
+            array("@var integer\n", "@var boolean\n", "@param integer\n", "@param boolean\n", "return integer\n", "return boolean\n"),
+            array("@var int\n", "@var bool\n", "@param int\n", "@param bool\n", "return int\n", "return bool\n"),
```
xabbuh: Why didn't you put a `@` in front of `return`?

phansys: I don't know @xabbuh :P (copy - paste typo). Fixed.

javiereguiluz:
Thanks @phansys.

Issue #9, opened 2011-06-23
Report added/modified/deleted files


lmcd:
I love the convenience of things like the interactive generator, but have a little anxiety/distrust for what they are doing behind the scenes. It always puts my mind at ease when I can see which files were added, modified or deleted.

Perhaps we should add this to the terminal output when running the generators?

javiereguiluz:
Is this feature still desired? It should not be that hard to implement it.

weaverryan:
@javiereguiluz I don't agree that it should require a `-v` flag. Beginners probably need to know what files are modified more than anyone, and they won't know about the `-v` flag (nor could they simply re-run the command after with `-v`).

PR #397, opened 2015-07-21
Report any file or directory created/deleted/changed

javiereguiluz:
This is a proof-of-concept PR to solve #9. The idea is to report any filesystem change with the labels `[+file]`, `[-file]`, `[+dir]`, `[-dir]` and `[changed]`.

For now it's only implemented in the `generate:bundle` command and it looks like the following (there are some `\n` issues pending to fix):

generate_bundle_report

### DIFF
--- a/Generator/Generator.php
+++ b/Generator/Generator.php
@@ -23,11 +23,6 @@ class Generator
     private $skeletonDirs;
     private static $output;
 
-    public function __construct()
-    {
-        self::$output = new ConsoleOutput();
-    }
-
     /**
      * Sets an array of directories to look for templates.
      *
@@ -75,18 +70,27 @@ class Generator
     public static function mkdir($dir, $mode = 0777, $recursive = true)
     {
         mkdir($dir, $mode, $recursive);
-        self::$output->writeln(sprintf('  <fg=green>created</> %s', self::relativizePath($dir)));
+        self::writeln(sprintf('  <fg=green>created</> %s', self::relativizePath($dir)));
     }
 
     public static function dump($filename, $content)
     {
         if (file_exists($filename)) {
-            self::$output->writeln(sprintf('  <fg=yellow>updated</> %s', self::relativizePath($filename)));
+            self::writeln(sprintf('  <fg=yellow>updated</> %s', self::relativizePath($filename)));
         } else {
-            self::$output->writeln(sprintf('  <fg=green>created</> %s', self::relativizePath($filename)));
+            self::writeln(sprintf('  <fg=green>created</> %s', self::relativizePath($filename)));
         }
 
         return file_put_contents($filename, $content);
+    }
+
+    private static function writeln($message)
+    {
+        if (null === self::$output = new ConsoleOutput()) {
+            self::$output = new ConsoleOutput();
+        }
+
+        self::output->writeln($message);
     }
 
     private static function relativizePath($absolutePath)
--- a/Generator/Generator.php
+++ b/Generator/Generator.php
@@ -86,11 +86,11 @@ class Generator
 
     private static function writeln($message)
     {
-        if (null === self::$output = new ConsoleOutput()) {
+        if (null === self::$output) {
             self::$output = new ConsoleOutput();
         }
 
-        self::output->writeln($message);
+        self::$output->writeln($message);
     }
 
     private static function relativizePath($absolutePath)
--- a/Generator/Generator.php
+++ b/Generator/Generator.php
@@ -75,22 +75,24 @@ class Generator
     public static function mkdir($dir, $mode = 0777, $recursive = true)
     {
         mkdir($dir, $mode, $recursive);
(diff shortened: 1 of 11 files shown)


hhamon:
I don't know if it makes sense or not but I would display the list of files when the command is run with a verbose option.

javiereguiluz:
@hhamon it makes sense of course ... but for others don't: Ryan seems to disagree with this: https://github.com/sensiolabs/SensioGeneratorBundle/issues/9#issuecomment-122902901

hhamon:
Thanks! This is what I was thinking about. For a newcomer I agree showing the list of files that have been created or changed is better.

javiereguiluz:
I've just updated this PR. The main change is that the labels are now `created`, `updated`, `deleted`, which are much more easy to understand. An example of the `generate:bundle` output:

generate_bundle_command

The proof-of-concept is now finished and it's time to make a decision: should we implement this idea for the rest of commands? Any changes that you want to suggest for the implementation? Thanks!

fabpot:
I'm a bit worried about the verbosity of the command. People don't read so many lines of output. Do you what is the max number of files that could be displayed by a command?

javiereguiluz:
The longest output would be to generate a shared bundle:

generate_shared_bundle

`generate:crud`, `generate:doctrine:entity`, `generate:doctrine:form`, etc. result in short outputs.

fabpot:
For a bundle generation, can't we just say that we generated output in the `src/Acme/BlogBundle/` directory? We don't really care about the individual files here as we generated the whole directory content.

javiereguiluz:
We need to make a decision about this change. There are 3 options:

1. Report ALL changes, as requested originally in #9 and as done by Rails, npm, etc.
2. Report MOST changes (as Fabien proposed, let's hide some changes when the list is long, like when creating a bundle)
3. Report NOTHING, like we do currently and close this pull request.

Let's ping some friends to help us decide: @weaverryan @WouterJ @xabbuh @stof @hhamon @phansys @OskarStark

OskarStark:
:+1:  for **Option 2** if we just chose one option

what about?
```default``` -> **Option 1**, so leave as it is
```-vv``` -> **Option 2**
```-vvv```-> **Option 3**

**Comment by stof** [comment]
showing this only for verbose output makes sense

**Comment by weaverryan** [comment]
My vote is for Option 1. Option 2 is also ok, but it seems like it might be a bit extra work, and I'm not convinced it's better. We generally don't create a lot of files (your screenshot is probably of our biggest operation), so I don't think it's overwhelming: it's informational!

**Comment by WouterJ** [comment]
I would either show the directory that's created without the files inside it (option 2) or the files that are created and not the directories. Showing both seem redundant to me (less output is better imo, especially for Windows users who have a very small cmd window by default).

**Comment by xabbuh** [comment]
I'm for option one too. Displaying messages depending on the output verbosity as proposed by @OskarStark even is a good idea, but probably is not worth the additional workload.

**Comment by fabpot** [comment]
Can you rebase? Is it mergeable?

**Comment by OskarStark** [comment]
@javiereguiluz can you please rebase this PR?

Thank you :+1:

**Comment by OskarStark** [comment]
ping @javiereguiluz

**Comment by fabpot** [comment]
@javiereguiluz What's the status of this PR?

**Comment by phansys** [comment]
@javiereguiluz, thank you so much for the call and sorry for the delay.
I agree with the solution of showing these messages based on the command's verbosity. Regarding the @weaverryan's [comment at #9](https://github.com/sensiolabs/SensioGeneratorBundle/issues/9#issuecomment-122902901) I think instead of polluting the default output with this kind of information, we should encourage beginners to execute the whole app (console commands too) in debug / verbose mode until they feel comfortable and familiar with the behavior. So, that way we are taking into account both scenarios while leveraging the current implementation for these kind of requirements.

File: Generator/Generator.php [hunk 1]
```diff
@@ -58,9 +61,42 @@ protected function getTwigEnvironment()
...
     {
         if (!is_dir(dirname($target))) {
-            mkdir(dirname($target), 0777, true);
+            self::mkdir(dirname($target));
+        }
+
+        return self::dump($target, $this->render($template, $parameters));
+    }
+
+    public static function mkdir($dir, $mode = 0777, $recursive = true)
```
HeahDude: could you please mark this method and the following `dump` as internal?

Manipulator/KernelManipulator.php
```diff
@@ -103,7 +104,7 @@ public function addBundle($bundle)
                 // remove last close parentheses
                 $leadingContent = rtrim(preg_replace($closingSymbolRegex, '', rtrim($leadingContent)));
 
-                if (substr($leadingContent, -1) !== '(' && substr($leadingContent, -1) !== '[' ) {
+                if (substr($leadingContent, -1) !== '(' && substr($leadingContent, -1) !== '[') {
```
HeahDude: can you please use yoda style here?

fabpot:
Thank you @javiereguiluz.

PR #401, opened 2015-08-10
Pass field mapping metadata to "new" view

javiereguiluz:
We already pass this information to the rest of the views (edit, show, etc.)


nicodmf:
Good news this is the last we waited, i thinks. Hoping this will be merged soon

PR #402, opened 2015-08-11
Validate the given bundle + entity name before generating the bundle

### LINKED ISSUE #338: Check if Entity namespace exists after giving shortcut name
Hi, I was generating CRUD entities with doctrine:generate:crud but I mistyped the shortcut name, typed Linda instead Lynda.
Wouldn't it be good to check namespace after giving the shortcut name?

screen shot 2015-03-09 at 10 54 50 am

javiereguiluz:
This fixes #338.


OskarStark:
:+1:

Issue #309, opened 2014-10-29
Add `Configuration.php` to resources for Container


DracoBlue:
Currently the generated Configuration.php wouldn't refresh the cache if it's modified in dev environment.

But if one adds

``` php
        $container->addResource(new FileResource(__DIR__.'/Configuration.php'));
```

in `DependencyInjection/*Extension.php` even modifications of the `DependencyInjection/Configuration.php` will trigger a cache refresh.

What do you think?

javiereguiluz:
Any other comment about this proposal?

javiereguiluz:
Closing it as "won't fix" for the reasons explained in #412.

PR #412, opened 2015-08-25
Added DI/Configuration to the resources watched by Config

javiereguiluz:
This implements #309.


xabbuh:
Imo this does not make much sense. Typically, you will add files/directories that contain the actual configuration of your application as container resources, but you don't do that with a class that merely defines the structure of the configuration.

javiereguiluz:
@xabbuh thanks for your comment. Anyone else wants to comment on this before merging or deleting it?

javiereguiluz:
Closing it because of the reasons stated by @xabbuh.

PR #413, opened 2015-09-02
Use PHP 7.0 instead of nightly

XWB:


javiereguiluz:
@XWB sorry for the slow response. Merged! Thanks.

PR #377, opened 2015-07-09
CRUD-generator follows symfony best practices

maidmaid:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #344
| License       | MIT
| Doc PR        |

Follow best practices :
- [x] [So it's much simpler to let newAction handle everything.](http://symfony.com/doc/current/best_practices/forms.html#handling-form-submits)
- [x] [Add buttons in the templates, not in the form classes or the controllers.](http://symfony.com/doc/current/best_practices/forms.html#form-button-configuration)
- [x] [Store all your application's templates in ``app/Resources/views/ directory``.](http://symfony.com/doc/current/best_practices/templates.html#template-locations)
- [x] [Rendering the Form](http://symfony.com/doc/current/best_practices/forms.html#rendering-the-form)
- [x] [We recommend using ``$form->isSubmitted()`` in the if statement for clarity.](http://symfony.com/doc/current/best_practices/forms.html#handling-form-submits)
- [x] [Use the ``ParamConverter`` trick to automatically query for Doctrine entities when it's simple and convenient.](http://symfony.com/doc/current/best_practices/controllers.html#using-the-paramconverter)
- [x]  [Use lowercased snake_case for directory and template names.](http://symfony.com/doc/current/best_practices/templates.html#template-locations)

Others :
- [x] Use ``redirectToRoute()`` shortcut notation for routing.

For see quickly the generated result, click on [this repo](https://github.com/maidmaid/sf-crud) ([controller](https://github.com/maidmaid/sf-crud/blob/master/src/AppBundle/Controller/ArticleController.php), [views](https://github.com/maidmaid/sf-crud/tree/master/app/Resources/views/article), [routing](https://github.com/maidmaid/sf-crud/tree/master/src/AppBundle/Resources/config/routing)).

PR #418, opened 2015-09-29
Load the main bundle's routing file form app's routing.yml file

javiereguiluz:
This fixes the problem described by @weaverryan in [this comment](https://github.com/sensiolabs/SensioGeneratorBundle/pull/377#issuecomment-124860772).

The error which will be usually displayed:

generator_1

The error when something really bad happens:

generator_2

Don't look at the style of the errors because I'll change everything soon. But please, review the actual content of the error messages. Thanks.

### DIFF
--- a/Command/GenerateDoctrineCrudCommand.php
+++ b/Command/GenerateDoctrineCrudCommand.php
@@ -267,7 +267,7 @@ EOT
         // second, import the bundle's routing.yml file from the application's routing.yml file
         $routing = new RoutingManipulator($this->getContainer()->getParameter('kernel.root_dir').'/config/routing.yml');
         try {
-            $ret = $auto ? $routing->addResource($bundle->getName(), $format) : false;
+            $ret = $auto ? $routing->addResource($bundle->getName(), 'yml') : false;
         } catch (\RuntimeException $e) {
             // the bundle is already imported form app's routing.yml file
             $errorMessage = sprintf(
--- a/Command/GenerateDoctrineCrudCommand.php
+++ b/Command/GenerateDoctrineCrudCommand.php
@@ -241,6 +241,8 @@ EOT
 
         $output->write('Importing the CRUD routes: ');
         $this->getContainer()->get('filesystem')->mkdir($bundle->getPath().'/Resources/config/');
+
+        // first, import the routing file from the bundle's main routing.yml file
         $routing = new RoutingManipulator($bundle->getPath().'/Resources/config/routing.yml');
         try {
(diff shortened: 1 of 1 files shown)


Command/GenerateDoctrineCrudCommand.php
```diff
@@ -261,6 +263,37 @@ protected function updateRouting(QuestionHelper $questionHelper, InputInterface
                 '',
             );
         }
+
+        // second, import the bundle's routing.yml file from the application's routing.yml file
+        $routing = new RoutingManipulator($this->getContainer()->getParameter('kernel.root_dir').'/config/routing.yml');
+        try {
+            $ret = $auto ? $routing->addResource($bundle->getName(), $format) : false;
```
weaverryan: the `$format` here should always be `yml`, correct? Because we want to import `@YourBundle/Resources/config/routing.yml`?

Command/GenerateDoctrineCrudCommand.php
```diff
@@ -261,6 +263,37 @@ protected function updateRouting(QuestionHelper $questionHelper, InputInterface
...
+        } catch (\RuntimeException $e) {
+            // the bundle is already imported form app's routing.yml file
+            $errorMessage = sprintf(
+                "\n\n[ERROR] The bundle's \"Resources/config/routing.yml\" file cannot be imported\n".
+                "from \"app/config/routing.yml\" because the \"%s\" bundle is\n".
+                "already imported. Make sure you are not using two different\n".
+                "configuration/routing formats in the same bundle because it won't work.\n",
+                $bundle->getName()
+            );
```
weaverryan: if I already have the import to `@AppBundle/Resources/config/routing.yml` in `app/config/routing.yml`, will this error be shown?

weaverryan:
@javiereguiluz it feels like we need a battery of functional tests for this bundle to make sure the edge-cases are taken care of in a decent way. There's a lot of manual testing currently.

Command/GenerateDoctrineCrudCommand.php
javiereguiluz: Yes, because the check made by RoutingManipulator is very simple: https://github.com/sensiolabs/SensioGeneratorBundle/blob/master/Manipulator/RoutingManipulator.php#L55-L60

javiereguiluz:
@weaverryan I agree completely. And we'd also need to refactor the entire bundle because most of its code is more complex than it should be.

Command/GenerateDoctrineCrudCommand.php
javiereguiluz: Good catch! Fixed. Thanks.

fabpot:
What's the status of this PR?

fabpot:
Thank you @javiereguiluz.

PR #421, opened 2015-10-02
use twig comment

OskarStark:


OskarStark:
ping @javiereguiluz

javiereguiluz:
Although it seems wrong, I think that the current example is right. Keep in mind that we are using a Twig template to generate PHP code. In fact, we are generating the PHPdoc section of a PHP file, so the PHP comments are right.

OskarStark:
I'm sorry you are right, i missed that...

Thank you!

PR #422, opened 2015-10-14
fix visibility of setFormat method

razmo:
This PR change the visibility of setFormat method.
This change is necessary if we need to extend the DoctrineCrudGenerator class.
If we redefine the generate method in our new class, we have an error because we call the setFormat method who is private.
The first solution is to change the visibility of setFormat method (that what i propose in this PR). The second solution is to duplicate the method in our new class (bad idea if we don't need to redefine it)

### DIFF
--- a/Generator/DoctrineCrudGenerator.php
+++ b/Generator/DoctrineCrudGenerator.php
@@ -107,7 +107,7 @@ class DoctrineCrudGenerator extends Generator
      *
      * @param string $format The configuration format
      */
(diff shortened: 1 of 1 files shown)


OskarStark:
looks OK to me

javiereguiluz:
It looks OK to me. I'll wait a bit more beore merging it in case someone wants to rais questions or concerns.

fabpot:
Thank you @razmo.

PR #425, opened 2015-11-05
Make this bundle work with symfony3

benji07:
I update the composer.json to allow installation with symfony3
I remove the getName on the FormType if it's not required (SF < 2.8)

I update the travis configuration to run the test with the last stable (2.7) and the 2 upcoming version (2.8 & 3.0)


Generator/DoctrineFormGenerator.php
```diff
@@ -88,6 +89,7 @@ public function generate(BundleInterface $bundle, $entity, ClassMetadataInfo $me
             // Add 'setDefaultOptions' method with deprecated type hint, if the new 'configureOptions' isn't available.
             // Required as long as Symfony 2.6 is supported.
             'configure_options_available' => method_exists('Symfony\Component\Form\AbstractType', 'configureOptions'),
+            'get_name_required' => Kernel::MAJOR_VERSION == 2 && Kernel::MINOR_VERSION < 8,
```
xabbuh: I would check for existence of the `getBlockPrefix()` method in the `AbstractType` class instead (version of the HttpKernel and Form component can differ).

Tests/Generator/DoctrineFormGeneratorTest.php
```diff
@@ -28,7 +29,14 @@ public function testGenerate()
         $this->assertContains('->add(\'updatedAt\', \'datetime\')', $content);
         $this->assertContains('class PostType extends AbstractType', $content);
         $this->assertContains("'data_class' => 'Foo\BarBundle\Entity\Post'", $content);
-        $this->assertContains("'foo_barbundle_post'", $content);
+
+        if (Kernel::MAJOR_VERSION == 2 && Kernel::MINOR_VERSION < 8) {
```
xabbuh: same here

benji07:
@xabbuh done

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: php
 
+php: [5.3.3, 5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
```
xabbuh: Is this change really needed? Imho it makes the diff bigger than necessary and makes it harder in the future to track when a really necessary change was done.

.travis.yml
```diff
@@ -10,11 +12,12 @@ install:
...
         - php: 5.6
-        - php: 7.0
-        - php: hhvm
+          env: DEPENDENCIES='dev' SYMFONY_VERSION="2.8.*@dev"
+        - php: 5.6
+          env: DEPENDENCIES='dev' SYMFONY_VERSION="3.0.*@dev"
     fast_finish: true
+
+before_script:
+    - if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
```
xabbuh: I don't think this is necessary when you use the stability flag with the Symfony version.

.travis.yml
benji07: no, i will revert that

benji07:
@xabbuh i fix the last 2 mistake, if everything is ok, i will squash / rebase my PR

fabpot:
Thank you @benji07.

fabpot:
I'm going to open another PR to slightly change the Travis strategy to be more in line with our best practices (for instance, we still want to test the minimum version of supported Symfony, which is not the case here).

benji07:
:+1:

Issue #426, opened 2015-11-07
Bundle enable in AppKernel fails with short array syntax


spinx:
This issue is present in master.

I use short array syntax like so:
![short syntax](http://i.imgur.com/pAzZri1.png)

After I generate the TestVendorTestBundle, I get this:
![failed injection](http://i.imgur.com/xtSaUZz.png)

PR #427, opened 2015-11-07
Fixed KernelManipulator to work with PHP short array syntax.

spinx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #426
| License       | MIT
| Doc PR        | ?

The code in KernelManipulator does account for short array syntax but it's not working. 

I added a fix and added all test cases for short syntax - test could probably be done better, please advise.

### DIFF
--- a/Tests/Manipulator/KernelManipulatorTest.php
+++ b/Tests/Manipulator/KernelManipulatorTest.php
@@ -60,13 +60,25 @@ class KernelManipulatorTest extends GeneratorTest
      */
     public function kernelStubFilenamesProvider()
     {
-        return array(
+        $stubs = array(
             'With empty bundles array' => array(__DIR__.'/Stubs/EmptyBundlesKernelStub.php'),
             'With empty multiline bundles array' => array(__DIR__.'/Stubs/EmptyBundlesMultilineKernelStub.php'),
             'With bundles array contains comma' => array(__DIR__.'/Stubs/ContainsCommaKernelStub.php'),
             'With bundles added w/o trailing comma' => array(__DIR__.'/Stubs/ContainsBundlesKernelStub.php'),
             'With some extra code and bad formatted' => array(__DIR__.'/Stubs/ContainsExtraCodeKernelStub.php'),
+
         );
+
+        if(PHP_VERSION_ID >= 50400){
+            $stubs = array_merge($stubs, array(
+                'With empty bundles array, short array syntax' => array(__DIR__.'/Stubs/EmptyBundlesShortArraySyntaxKernelStub.php'),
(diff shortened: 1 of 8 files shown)


spinx:
The build is failing on PHP 5.3 which is a non-issue, right ?

xabbuh:
You need to skip the short array syntax tests on PHP 5.3.

spinx:
@xabbuh Yeah, I realised I could do that. I need to setup PHP 5.3 somewhere to test the fix that I did. 

BUT since this is in master, I was assuming PHP 5.3 support is no longer needed. My first contribution to SF2 ecosystem so please bear with me - where are the supported php versions specified ?

javiereguiluz:
@spinx thanks for your first contribution!  There is no need for you to setup a 5.3 environment. We use Travis to test code in different setups (https://travis-ci.org/sensiolabs/SensioGeneratorBundle/builds/89845420) This is precisely the error mentioned by @xabbuh that happened in 5.3 version: https://travis-ci.org/sensiolabs/SensioGeneratorBundle/jobs/89845422

You just need to check if PHP version is less than 5.4.0 and in that case, mark the test as skipped. Thanks.

spinx:
@javiereguiluz Tests pass, squashed commits.

I don't pass stubs if less than PHP 5.3 since these tests are doing some magic and I don't see how could I skip them in a way that would be acceptable.

Tests/Manipulator/KernelManipulatorTest.php
```diff
@@ -60,13 +60,25 @@ public function testAddToArray($kernelOriginFilePath)
...
+        $stubs = array(
             'With empty bundles array' => array(__DIR__.'/Stubs/EmptyBundlesKernelStub.php'),
             'With empty multiline bundles array' => array(__DIR__.'/Stubs/EmptyBundlesMultilineKernelStub.php'),
             'With bundles array contains comma' => array(__DIR__.'/Stubs/ContainsCommaKernelStub.php'),
             'With bundles added w/o trailing comma' => array(__DIR__.'/Stubs/ContainsBundlesKernelStub.php'),
             'With some extra code and bad formatted' => array(__DIR__.'/Stubs/ContainsExtraCodeKernelStub.php'),
+
         );
+
+        if(version_compare(PHP_VERSION, '5.4.0') >= 0){
```
xabbuh: In Symfony we usually use the `PHP_VERSION_ID` constant to do checks like this:

```php
if (PHP_VERSION_ID >= 50400) {
    // ...
}

spinx: Fixed.

fabpot:
Thank you @spinx.

PR #430, opened 2015-11-19
Added a period at the end of the sentences to conform to Symfony specifications.

jaytaph:
When running `php-cs-fixer` over Symfony generated bundles, it will find and correct two docblock sentences by adding periods to the end. This PR will automatically add these periods to the bundle skeleton, so at least the bundle generator generates code that conforms fully to the Symfony specification / php-cs-fixer.

### DIFF
--- a/Resources/skeleton/bundle/Configuration.php.twig
+++ b/Resources/skeleton/bundle/Configuration.php.twig
@@ -9,7 +9,7 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 
 /**
 {% block phpdoc_class_header %}
- * This is the class that validates and merges configuration from your app/config files
+ * This is the class that validates and merges configuration from your app/config files.
 {% endblock phpdoc_class_header %}
  *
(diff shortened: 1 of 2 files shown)


OskarStark:
:+1: for your contribution

can you please check why travis failed?

Thank you

jaytaph:
@OskarStark I've taken a look, and it seems to be an issue with the PHP finder, not with this code.. I'll try and pinpoint how and why (it seems that finding PHP binaries has changed, so it will include an additional space after the binary name..  running the process would use the arguments like: `'/usr/bin/php<space>', '-l', '...'`..  meaning that the binary cannot be found.

jaytaph:
The process issue has been fixed in symfony 2.3, just waiting for the merge towards 2.7, 2.8 and master. Until then, I reckon each PR will fail until then.

OskarStark:
:+1: @jaytaph

fabpot:
Thank you @jaytaph.

Issue #435, opened 2015-12-01
Doctrine:generate:crud doesn't output valid symfony3 form 


FabienPapet:
Doctrine:generate:crud doesn't output valid symfony3 controller which leads to an `UnexpectedTypeException`.

Example : for an entity named `ModuleModel`

### Existing
``` 
<?php 
class ModuleModelController {
    public function fooAction(Request $request)
    {
        [...]
        $form = $this->createForm(new ModuleModelType(), $moduleModel);
        [...]
    }
}
```

### Expected

``` 
<?php 
class ModuleModelController {
    public function fooAction(Request $request)
    {
        [...]
        $form = $this->createForm(ModuleModelType::class, $moduleModel);
        [...]
    }
}
```

<hr>

Neither for generated formtype

PR #436, opened 2015-12-01
#435 Fixed actions and views

jpyzio:
After resolve action problems, there was problems in views, crud generete for example:

In client/index.html.twig:
```
<td>{% if entity.isIndividual %}Yes{% else %}No{% endif %}</td>
```
insted:
```
<td>{% if client.isIndividual %}Yes{% else %}No{% endif %}</td>
```

It was fixed too in commit.


Resources/skeleton/crud/actions/edit.php.twig
```diff
@@ -17,7 +17,7 @@
     {
 {% block method_body %}
         $deleteForm = $this->createDeleteForm(${{ entity_singularized }});
-        $editForm = $this->createForm(new {{ entity_class }}Type(), ${{ entity_singularized }});
+        $editForm = $this->createForm({{ entity_class }}Type::class, ${{ entity_singularized }});
```
xabbuh: Actually, this does not work in PHP versions before 5.5. To be on the safe side you should probably rather use the FCQN instead.

OskarStark: :+1: for FQN

OskarStark: maybe a check if php > 5.3 is installed or something like this or an option in the commands?

Command/GenerateCommandCommand.php
```diff
@@ -43,7 +43,7 @@ public function configure()
 inside bundles. Provide the bundle name as the first argument and the command
 name as the second argument:
 
-<info>php app/console generate:command AppBundle blog:publish-posts</info>
+<info>php bin/console generate:command AppBundle blog:publish-posts</info>
```
fabpot: All the `app/` to `bin/` changes must be reverted as we support both 2.x and 3.x dir structure.

jpyzio: Done.

khalid-s:
hi, just gave a try for this solution, and works great.

But when generating a crud for a bundle living in a vendor name, there is a \ missing between the vendor name and the bundle name for new and edit action. not sure that am clear. e.g

For a bundle CompanyBundle in the folder  /src/IKNSA/CompanyBundle/Controller/CompanyController.php 

```
public function newAction(Request $request)
    {
        $company = new Company();
        $form = $this->createForm('\IKNSACompanyBundle\Form\CompanyType', $company);
        $form->handleRequest($request);
```
There is a backslash missing between IKNSA and CompanyBundle

Resources/skeleton/crud/actions/edit.php.twig
```diff
@@ -17,7 +17,7 @@
     {
 {% block method_body %}
         $deleteForm = $this->createDeleteForm(${{ entity_singularized }});
-        $editForm = $this->createForm(new {{ entity_class }}Type(), ${{ entity_singularized }});
+        $editForm = $this->createForm('\{{ bundle }}\Form\{{ entity_class }}Type', ${{ entity_singularized }});
```
fabpot: The leading `\` is not needed, same below.

fabpot: IIRC, `bundle` references the bundle name, not the bundle namespace.

jpyzio:
It should work fine.

khalid-s:
Tested again and it works fine

fabpot:
Thank you @jpyzio.

PR #437, opened 2015-12-02
Updating app/console to bin/console

coreight:


xabbuh:
see also #438 and #442

fabpot:
see #442

PR #441, opened 2015-12-05
Fix precision parameter handling.

AdrienLucas:
It also remove an unused variable some lines above (not sure if it worth having its own PR).


xabbuh:
:+1:

fabpot:
Thank you @AdrienLucas.

OskarStark:
good catch :+1:

PR #438, opened 2015-12-03
Update generate_bundle.rst

dzordzan:
Updating console commands to symfony 3

PR #442, opened 2015-12-07
fixed console path

milosmirkovic:
Console has been moved from app to bin folder.


fabpot:
The bundle works with Symfony 2.7, 2.8, and 3.0. So, `app/console` and `bin/console` are right :)

xabbuh:
see also #437 and #438

milosmirkovic:
@fabpot After installing Symfony 3.0.0 using Symfony Installer, command  'php app/console generate:bundle' will respond with 'Could not open input file: app/console' error.

PR #444, opened 2015-12-07
Update generate_bundle.rst

mrsuh:
i install symfony 3.0 and script "console" locate in /bin folder now


fabpot:
duplicates of #442

PR #440, opened 2015-12-05
Fixing typos

AdrienLucas:
Just fixing typos.

PR #445, opened 2015-12-07
fix argument type PHPdoc

xabbuh:
see https://github.com/sensiolabs/SensioGeneratorBundle/pull/440#discussion_r46885848


fabpot:
Thank you @xabbuh.

PR #446, opened 2015-12-09
Fix Twig index and show template entity name insertion for array and boolean type

jckeen:
When an entity metadata.type was an array or boolean in show.html.twig.twig or index.html.twig.twig, it would not utilize the actual entity name, instead it would use 'entity'.

This would throw a Twig_Error_Runtime:

> Variable "entity" does not exist in review\index.html.twig at line 25

This pull request fixes the problem.


xabbuh:
This is already covered by #436.

jckeen:
Sorry I missed that. Thanks.

PR #450, opened 2015-12-11
Fixed entity object name in newAction

DerekRoth:
Object name in newAction was using the lowercase class name instead of singularized entity name.


kissge:
Should be merged as early as possible! I nearly sent the same pull request.

fabpot:
Thank you @DerekRoth.

PR #451, opened 2015-12-14
Fix can't abort generate:controller

kenjis:
Now we can't abort controller generation.

~~~
...
You are going to generate a "AppBundle:Post" controller
using the "annotation" format for the routing and the "twig" format
for templating
Do you confirm generation [yes]? no

                         
  Controller generation  
                         

Generating the bundle code: OK

                                         
  Everything is OK! Now get to work :).  
~~~


fabpot:
Thank you @kenjis.

PR #452, opened 2015-12-17
Console resides on bin folder

ahmadsajid1989:
I have updated the console commands for symfony 3 version, from now, the console resides into bin folder but in documentation its written app, which is valid for previous version.


xabbuh:
It's still valid as this bundle still supports Symfony 2.x.

fabpot:
Both are valid.

PR #454, opened 2015-12-18
Fix PSR-2 line length recommendations

eliecharra:
Extension template is generating an inline phpdoc `@link` which exceed PSR-2 line lenght recommendation.

http://www.php-fig.org/psr/psr-2/#1-overview

> There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less.

`@link` anotation is enough imho, maybe previous text could be removed to get under recommended line length. 

 Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


fabpot:
Thank you @eliecharra.

eliecharra:
:+1:

PR #455, opened 2015-12-20
Allows digits in validateEntityName

intermike:


OskarStark:
Could you please add a test for this change?

intermike:
I am not sure I did it properly and fully. Please check. Thanks.

intermike:
Btw, I did this new test without primal change and phpunit has been just stuck without any error messages. Is that normal behaviour?

clipboard01

javiereguiluz:
@intermike after applying your change tests work for me (and for Travis CI service too):

phpunit_tests

So I'm going to merge this change. Thanks for your contribution.

PR #458, opened 2015-12-25
Update generate_bundle.rst

DavyGab:


OskarStark:
:-1:

DavyGab:
??
With Symfony 3, it's not app/console !

xabbuh:
That's true, but for Symfony 2 it's still `app/console` and this bundle works with both versions.

DavyGab:
Ok. I thought I changed only for Symfony 3... Because I lost time to "find solution".

fabpot:
Closing as there is no good way to fix that.

PR #456, opened 2015-12-20
Replace app/console by bin/console

pixeltoo:
Hi,
In the version 3.0 of Symfony2 the command line is php bin/console not php app/console
See the answer of this post http://stackoverflow.com/questions/34045622/could-not-open-input-file-app-console-symfony-2-in-both-mamp-and-vagrant
Best regards.

PR #460, opened 2015-12-28
Added caution notes for Symfony 3 apps

javiereguiluz:
This proposal is an alternative way of fixing the problems explained in #458 and #456.


xabbuh:
You could create a single file for this note and include it where needed.

fabpot:
Thank you @javiereguiluz.

PR #461, opened 2016-01-01
Implement quoted services references for Symfony 3

gabiudrescu:
I think by using Gherkin is the easiest to describe the scope of this pull request:

Given the symfony instance is 2.8 or upper
If the service has a dependency on another service
Then the service argument should be quoted in "
And no deprecation notice will be raised

If the generator is currently used to create a bundle on a 2.8 / 3 instance of Symfony, the services.yml file will have examples of services without quotes.

E.g:
```
services:
#    game.example:
#        class: GameBundle\Example
#        arguments: [@service_id, "plain_value", %parameter%]
```

To avoid potential troubles, I thought the example should be as I have modified it with my pull request.

### DIFF
--- a/Resources/skeleton/bundle/services.yml.twig
+++ b/Resources/skeleton/bundle/services.yml.twig
@@ -2,5 +2,5 @@ services:
 {% block services %}
 #    {{ extension_alias }}.example:
 #        class: {{ namespace }}\Example
-#        arguments: [@service_id, "plain_value", %parameter%]
+#        arguments: ["@service_id", "plain_value", %parameter%]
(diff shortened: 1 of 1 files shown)


xabbuh:
I think we should simply always removed single quotes instead of double quotes (in double quotes you need to be careful to not use escape sequences, for example).

gabiudrescu:
@xabbuh not a bad idea. I kinda followed the example in place right now, but if you or anyone else finds this a good idea, it can be modified.

fabpot:
Thank you @gabiudrescu.

PR #464, opened 2016-01-05
Fix #463 Form Generator don't use FieldType::class

### LINKED ISSUE #463: Form Generator don't use FieldType::class
When generate a Form and the Entity contain a field type in `[date,time,datetime]`, the FormType generated don't use `FieldType::class`.

```php
    public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder
            ->add('author')
            ->add('content')
            ->add('date', 'datetime')
            ->add('advert')
        ;
    }
```

Since Symfony 2.8

```php
    public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder
            ->add('author')
            ->add('content')
            ->add('date', 'Symfony\Component\Form\Extension\Core\Type\DateTimeType')
            ->add('advert')
        ;
    }
```

roromix:
This fixes #463.

### DIFF
--- a/Resources/skeleton/form/FormType.php.twig
+++ b/Resources/skeleton/form/FormType.php.twig
@@ -27,8 +27,13 @@ class {{ form_class }} extends AbstractType
 
         {%- for field in fields -%}
             {%- if fields_mapping[field] is defined and fields_mapping[field]['type'] in ['date', 'time', 'datetime'] %}
+                {%- if fields_mapping[field]['type'] == 'datetime' %}
+                    {%- set fieldType = 'DateTime' %}
+                {%- else %}
+                    {%- set fieldType = fields_mapping[field]['type']|capitalize %}
+                {%- endif %}
 
-            ->add('{{ field }}', '{{ fields_mapping[field]['type'] }}')
+            ->add('{{ field }}', 'Symfony\Component\Form\Extension\Core\Type\{{ fieldType }}Type')
 
             {%- else %}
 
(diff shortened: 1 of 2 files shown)


Tobion:
Please make it compatible with Symfony < 2.8 as well (where the named type still needs to be used). Several bundles already support that if you need an example (maybe also described in the symfony upgrade file).

chalasr:
:+1: Thanks for this.

Resources/skeleton/form/FormType.php.twig
```diff
@@ -27,8 +27,13 @@ class {{ form_class }} extends AbstractType
...
         {%- for field in fields -%}
             {%- if fields_mapping[field] is defined and fields_mapping[field]['type'] in ['date', 'time', 'datetime'] %}
+                {%- if fields_mapping[field]['type'] == 'datetime' %}
+                    {%- set fieldType = 'DateTime' %}
+                {%- else %}
+                    {%- set fieldType = fields_mapping[field]['type']|capitalize %}
+                {%- endif %}
 
-            ->add('{{ field }}', '{{ fields_mapping[field]['type'] }}')
+            ->add('{{ field }}', 'Symfony\Component\Form\Extension\Core\Type\{{ fieldType }}Type')
```
HeahDude: The fqcn for `entity` is `Symfony\Bridge\Doctrine\Form\Type\EntityType`, since it belongs to the core can you please add this case?

HeahDude:
Ok, I'm closing here in favor of #511, because we have to keep BC with 2.7. Thanks!

PR #468, opened 2016-02-04
Update generate_bundle.rst

milardovich:
Please change app/console for bin/console for 3.0 version


OskarStark:
:-1: please use the search functionality, this PR was proposed several times, we won't accept it.

maybe @xabbuh can link the other issues

javiereguiluz:
If one user asks something ... he may be wrong. If lots of users ask for the same thing ... we are probably wrong and need to fix something.

xabbuh:
Imo the note added at the top of all chapters (see #460) should be enough.

OskarStark:
:+1:

javiereguiluz:
@milardovich thanks for proposing this change. This issue is a bit tricky because the bundle is compatible both with Symfony 2 and 3. Right now we display a note about this at the top of each page:

generator_bundle

That's why I'm closing this as "won't fix" for now. But we'll keep thinking about ways to improve this. Thanks!

PR #471, opened 2016-02-09
Update generate_bundle.rst

BonnieDoug:
Old style app/console changed to bin/console.


BonnieDoug:
Hadn't read the comments to these same pull requests before adding this one. Closed for the reasons stated #460 #458 #456 

Not that I agree with the reasoning. Personally I think we should document for up to date versions and add a note to warn for legacy systems not the other way round.

PR #477, opened 2016-03-02
Use FormType::class notation instead of string constants to create forms

systemist:


xabbuh:
I am :-1: on this change as the `class` constant is not available in all PHP versions that the bundle supports.

javiereguiluz:
@systemist I like your proposal but I need to close it as "won't fix" because of the reasons given by @xabbuh. `::class` is only available starting from PHP 5.5.

PR #479, opened 2016-03-07
Resolves the issues of crud & entity generation when the entity is namespaced

starx:


fabpot:
Change of the execution flag on files should be reverted (0644 instead of 0755)

Resources/skeleton/crud/actions/edit.php.twig
```diff
@@ -17,7 +17,7 @@
     {
 {% block method_body %}
         $deleteForm = $this->createDeleteForm(${{ entity_singularized }});
-        $editForm = $this->createForm('{{ namespace }}\Form\{{ entity_class }}Type', ${{ entity_singularized }});
+        $editForm = $this->createForm('{{ namespace }}\Form\{% if entity_namespace is not empty %}{{ entity_namespace }}\{% endif %}{{ entity_class }}Type', ${{ entity_singularized }});
```
fabpot: Where does that `entity_namespace`variable come from?

starx: @fabpot On `generateControllerClass()` of `DoctrineCrudGenerator`.

starx:
@fabpot File mode have been updated.

HeahDude:
Thank you @starx for working on this :) However I'm gonna close here in favor of #508 and #509 which are a bit different and the most important add some tests to prevent regressions.

Thank you for understanding.

PR #489, opened 2016-04-15
resources views root path replace bundle path

gonzaloalonsod:
$this->rootDir change $this->bundle->getPath()

### DIFF
--- a/Generator/DoctrineCrudGenerator.php
+++ b/Generator/DoctrineCrudGenerator.php
@@ -78,7 +78,7 @@ class DoctrineCrudGenerator extends Generator
 
         $this->generateControllerClass($forceOverwrite);
 
-        $dir = sprintf('%s/Resources/views/%s', $this->rootDir, str_replace('\\', '/', strtolower($this->entity)));
+        $dir = sprintf('%s/Resources/views/%s', $this->bundle->getPath(), str_replace('\\', '/', strtolower($this->entity)));
 
         if (!file_exists($dir)) {
             $this->filesystem->mkdir($dir, 0777);


javiereguiluz:
@gonzakpo I'd say that `generate:crud` command is only used for application bundles (not third-party bundles). These app bundles store their templates in `app/Resources/views/` by default, so we should keep using `rootdir` instead of `bundle->getPath()`, right?

gonzakpo:
Yes. Personally I prefer to have the views in the bundle, as it was before. Any chance of settings?
For example ask if bundle (not third-party bundles) would be acceptable?

Generator/DoctrineCrudGenerator.php
```diff
@@ -78,7 +78,7 @@ public function generate(BundleInterface $bundle, $entity, ClassMetadataInfo $me
 
         $this->generateControllerClass($forceOverwrite);
 
-        $dir = sprintf('%s/Resources/views/%s', $this->rootDir, str_replace('\\', '/', strtolower($this->entity)));
+        $dir = sprintf('%s/Resources/views/%s', $this->bundle->getPath(), str_replace('\\', '/', strtolower($this->entity)));
```
omrqs: I was suggested this:
$dir = sprintf('%s/Resources/views/%s', $this->bundle->getPath(), str_replace('\\', '/', ucfirst($this->entity)));

xabbuh:
@javiereguiluz The `GenerateBundleCommand` has a `shared` option. Not sure if adding that was a good idea, but shouldn't the other command support the same for consistency?

fabpot:

PR #490, opened 2016-04-25
Use locateResource to get bundle's skeleton resources

robinvdvleuten:
Hi there,

I've replaced the hard-coded skeleton dir path to a locateResource() call so it's possible for a bundle to extend the templates.

Cheers,
Robin


javiereguiluz:
I'm 👍 ... but not strongly. I'm not sure if this feature will be used a lot.

fabpot:

PR #491, opened 2016-05-07
Update generate_doctrine_form.rst

jezemery:
These are Symfony 3 docs, therefore I feel they should target Symfony 3 first, and older versions second as most of the rest of the documentation seems to. Happy to have feedback!


javiereguiluz:
@jezemery thanks for proposing this update. I agree that it's time to prioritize Symfony 3.x. I've made some minor rewordings in 5696bff. Thanks.

PR #494, opened 2016-05-22
Generating the tests into the root tests/ directory for non-shared bundles

### LINKED ISSUE #416: Create the `Tests` directory elsewhere
According to symfony/symfony-standard#855 , the `{Bundle}/Tests` directory will be moved to the root `tests/` directory, in the Symfony 3 directory structure.

The problem is that the `generate:bundle` command still creates the `Tests` directory.

What can be done IMO: 

* Create a new `--tests-dir` option.<br>
    ```bash
    $ php bin/console generate:bundle -n --dir=src --namespace=Acme/MyFeatureBundle --bundle-name=AcmeMyFeatureBundle -n
    > Will create src/Acme/MyFeatureBundle/Tests normally
    $ php bin/console generate:bundle -n --dir=src --namespace=Acme/MyFeatureBundle --bundle-name=AcmeMyFeatureBundle --tests-dir=tests
    > Will create tests/AcmeMyFeatureBundle/ and put the default `Controller` directory inside
    ```
    This behavior sounds to me like the `app/Resources/{BundleName}/` structure, so it might be the most understandable to everyone, but of course it constraints the new directory to have the same name as the bundle.

* Add a new `symfony-tests-dir` parameter in the root `composer.json` file, and set it as `tests` by default, and get the information from the command to know where to dump the default tests files.

* Force the tests to be put inside the root `tests` directory and bump another major version of the bundle whenever Symfony3 releases.

What do you think?

weaverryan:
This fixes #416 in a very sensible way:

A) If the bundle is "shared", then the tests go inside the bundle (as always)
B) If the bundle is not shared, then the tests go into `/tests` at the root of the project.

I've tested this and it works great in both cases. I *am* using `getcwd()` to find the root of the project. But this is already done in the generator to find the `src/` directory. So if we're worried about this assumption, it should be fixed later.

Thanks!

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -386,8 +386,11 @@ EOT
         }
         $format = Validators::validateFormat($input->getOption('format'));
 
+        // an assumption that the kernel root dir is in a directory (like app/)
+        $projectRootDirectory = $this->getContainer()->getParameter('kernel.root_dir').'/..';
+
         if (!$this->getContainer()->get('filesystem')->isAbsolutePath($dir)) {
-            $dir = getcwd().'/'.$dir;
+            $dir = $projectRootDirectory.'/'.$dir;
         }
         // add trailing / if necessary
         $dir = '/' === substr($dir, -1, 1) ? $dir : $dir.'/';
@@ -402,7 +405,7 @@ EOT
 
         // not shared - put the tests in the root
         if (!$shared) {
-            $testsDir = getcwd().'/tests/'.$bundleName;
+            $testsDir = $projectRootDirectory.'/tests/'.$bundleName;
             $bundle->setTestsDirectory($testsDir);
         }
 
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -392,13 +392,21 @@ EOT
         // add trailing / if necessary
         $dir = '/' === substr($dir, -1, 1) ? $dir : $dir.'/';
 
-        return new Bundle(
+        $bundle = new Bundle(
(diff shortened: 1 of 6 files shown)


Pierstoval:
This `getcwd` is quite strange at first, but it works, thanks for the feature :+1: can't wait for it to be implemented :)

Command/GenerateBundleCommand.php
```diff
@@ -392,13 +392,21 @@ protected function createBundleObject(InputInterface $input)
...
             $namespace,
             $bundleName,
             $dir,
             $format,
             $shared
         );
+
+        // not shared - put the tests in the root
+        if (!$shared) {
+            $testsDir = getcwd().'/tests/'.$bundleName;
```
xabbuh: I would base this on the kernel's root directory instead.

fabpot: agreed

weaverryan:
Updated! This should be good to go now

Tests/Command/GenerateCommandTest.php
```diff
@@ -18,6 +18,13 @@
 
 abstract class GenerateCommandTest extends \PHPUnit_Framework_TestCase
 {
+    protected $kernelRootDir;
+
+    public function setup()
+    {
+        $this->kernelRootDir = sys_get_temp_dir();
```
xabbuh: I would generate a unique directory name and also clean it up after the tests have been executed.

weaverryan:
@xabbuh I just solved it in a different way that required less changes :)

fabpot:
Thank you @weaverryan.

PR #495, opened 2016-05-23
Full class name for form type

whyte624:


Resources/skeleton/form/FormType.php.twig
```diff
@@ -28,7 +28,7 @@ class {{ form_class }} extends AbstractType
         {%- for field in fields -%}
             {%- if fields_mapping[field] is defined and fields_mapping[field]['type'] in ['date', 'time', 'datetime'] %}
 
-            ->add('{{ field }}', '{{ fields_mapping[field]['type'] }}')
+            ->add('{{ field }}', 'Symfony\Component\Form\Extension\Core\Type\{{ fields_mapping[field]['type']|capitalize }}Type')
```
javiereguiluz: Sadly this won't work because form types contain some irregular namespaces. For example: `datetime` is not defined in `Symfony\...\DatetimeType` but `Symfony\...\DateTimeType`  And the `entity` type is defined in the `DoctrineBridge`  

Because of these inconsistencies, the only surefire way to do this mapping is to define the entire map your self. For example: https://github.com/javiereguiluz/EasyAdminBundle/blob/master/Form/Util/LegacyFormHelper.php#L23

whyte624: Thank you, for pointing this out!

whyte624:
Already implemented in https://github.com/sensiolabs/SensioGeneratorBundle/pull/464

PR #497, opened 2016-06-13
Replace getMock calls with getMockBuilder()->getMock() calls in tests

hairmare:
The upgrade to PHPUnit 5.4 wants us to use `createMock` instead of `getMock` as per [the PHPUnit 5.4 Releasenotes](https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-5.4.0#changes).

This PR changes the tests so they execute properly using the latest PHPUnit.

### DIFF
--- a/Tests/Command/AutoComplete/EntitiesAutoCompleterTest.php
+++ b/Tests/Command/AutoComplete/EntitiesAutoCompleterTest.php
@@ -66,14 +66,14 @@ class EntitiesAutoCompleterTest extends \PHPUnit_Framework_TestCase
      */
     protected function getEntityManagerMock($aliases, $classes)
     {
-        $cache = $this->getMock('Doctrine\Common\Persistence\Mapping\Driver\MappingDriver');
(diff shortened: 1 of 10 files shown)


xabbuh:
As you can see in the Travis build this fails on older PHPUnit versions. If you want to avoid the deprecated `getMock()` method here, I suggest to use `getMockBuilder()` instead which is available on those versions too.

hairmare:
Thanks @xabbuh I just noticed myself and will update the PR so it still works with older phpunit versions.

hairmare:
I updated this PR so it works with PHPUnit <5 as well as PHPUnit >= 5.4.

narcoticfresh:
@xabbuh can this be merged?

xabbuh:
@narcoticfresh Sorry, I am not a maintainer of this bundle. But the changes so far look good to me.

fabpot:
@hairmare Looks good to me as well. Small nitpicking: can you move the whole expression on one line for each call?

hairmare:
@fabpot I addressed the formatting changes and rebased the pull request. Thanks

fabpot:
Thank you @hairmare.

PR #498, opened 2016-06-13
Add bundles name auto-completion.

adrienlucas:
This adds the auto-completion of bundles name in the `generate:controller` and the `generate:command` commands.
This is only a lousy copy-pasting of the auto-completion made in the `generate:doctrine:crud` command.


javiereguiluz:
👍 

This looks OK to me and it's very useful. Thanks @adrienlucas.

fabpot:
Thank you @adrienlucas.

PR #503, opened 2016-08-10
Update generate_bindle.rst

CyrilCharlier:
In Symfony 3.x it is bin/console not app/console


fabpot:
The bundle is compatible with both versions of Symfony.

PR #504, opened 2016-08-12
Remove dev minimum stability

hason:


fabpot:
Why? That's what we are doing everywhere and it does not affect your own code anyway.

PR #506, opened 2016-08-18
Update generate_bundle.rst

lucchese:


Resources/doc/commands/generate_bundle.rst
```diff
@@ -3,7 +3,7 @@ Generating a New Bundle Skeleton
 
 .. caution::
 
-    If your application is based on Symfony 3, replace ``php app/console`` by
+    If your application is based on Symfony 3, replace ``php bin/console`` by
     ``php bin/console`` before executing any of the console commands included
```
fabpot: The current sentence is correct. If you use Symfony 3, you should use `bin/console` instead of `app/console` which is the path for Symfony 2.

Resources/doc/commands/generate_bundle.rst
```diff
@@ -19,14 +19,14 @@ structure:
 
 .. code-block:: bash
 
-    $ php app/console generate:bundle
+    $ php bin/console generate:bundle
```
fabpot: The comment above tells you to use bin instead of app for Symfony 3, so this is correct.

PR #507, opened 2016-08-27
Fix for Undefined variable: actionName error

johnnybenson:
Using generator with actions option fails

```sh
php bin/console generate:controller -v --controller=SomeBundle:Entity --actions="showAction:/path"
```

image

[Symfony\Component\Debug\Exception\ContextErrorException]
  Notice: Undefined variable: actionName


fabpot:
Thank you @johnnybenson.

Issue #465, opened 2016-01-05
Unable to generate a table from an entity contained in subfolder


simohammedhttp:
I tried to generate a `Client` entity:

```bash
$ php app/console generate:doctrine:entity --no-interaction
  --entity=MyBundle:Surveys\Client --fields="..." --format=annotation
```

Inside the generated entity, I got an unexpected table name :

```php
@ORM\Table(name="surveys\client")
```

The corresponding sql query:

```bash
$ php app\console doctrine:schema:update --dump-sql
```

```sql
CREATE TABLE surveys\client (id INT AUTO_INCREMENT NOT NULL, ....)
DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB
```

After forcing the query execution, I got an error :

```
check the manual that corresponds to your MySQL server version for the
right syntax to use near '\client
```

Executing the same query after wrapping the name of the table with sql quotes CREATE TABLE `surveys\client` works fine.

But after all, the "Client" entity is intended to be hosted under the subfolder "Surveys", I wonder if we can set automatically the correct name inside the generated entity : 

```php
@ORM\Table(name="client")
```

ju1ius:
The problem comes from the generated `@Table` annotation:
```sh
$ app/console generate:doctrine:entity --entity=AcmeBlogBundle:Blog/Post
```
```php
namespace Acme\BlogBundle\Entity\Blog;

/**
 * @ORM\Table(name="blog\post")
 * @ORM\Entity()
 */
class Post {}
```
One would expect something like `@ORM\Table(name="blog__post")`

I must also add that entities generated in subfolders are ignored by the schema tool.
After generating the above entity and fixing the table name:
```sh
$ app/console doctrine:schema:update
> Nothing to update - your database is already in sync with the current entity metadata.
```
:confused:

HeahDude:
Hi there, this issue comes from Doctrine ORM generator. If you still encounter this problem, please report it there. Thanks

PR #509, opened 2016-10-08
Fixed table name in subnamespaced entity annotations

HeahDude:
Closes https://github.com/sensiolabs/SensioGeneratorBundle/issues/465.

Looks like I was wrong I've been able to add a test to reproduce the issue.


fabpot:
Thank you @HeahDude.

PR #512, opened 2016-10-09
Fixed cs in manipulators

HeahDude:


fabpot:
Thank you @HeahDude.

Manipulator/KernelManipulator.php
```diff
@@ -23,11 +23,6 @@ class KernelManipulator extends Manipulator
     protected $kernel;
     protected $reflected;
 
-    /**
-     * Constructor.
-     *
-     * @param KernelInterface $kernel A KernelInterface instance
```
OskarStark: why removing this?

HeahDude: Does not bring any value, the signature is explicit enough

OskarStark: makes no senso to me, but I'm fine with this.

I was just interested in this, thanks for the feedback

PR #513, opened 2016-10-09
Fixed paths in crud controller annotation

HeahDude:
Route annotation at controller level generated by the crud command has two slashes `//post` or `//blog_post` for namespaced entities.


fabpot:
Thank you @HeahDude.

PR #514, opened 2016-10-09
[DX] Normalized entity argument for doctrine commands

HeahDude:
Follows up #280 and #286.

This was never documented. I suggest to deprecate the option as it is more verbose, and currently the only way for the form generation.


fabpot:
I'm not comfortable with this one. I'm pretty sure that there was a reason for having everything as an option (I unfortunately don't remember the exact one).

HeahDude:
@fabpot is this the fact that an argument cannot be required and resolved while interacting as options do.

fabpot:
Should be that indeed.

fabpot:
Thank you @HeahDude.

PR #517, opened 2016-10-19
Change route from "app" to "bin"

wilson-vargas:
When you write "app/console" returns: **Could not open input file: app/console**.

I changed this to **bin/console**


xabbuh:
`app/console` is still true for older versions of the Symfony Standard Edition (for that there's a hint noting that you have to replace `app/console` with `bin/console` on top of the page). See also https://github.com/sensiolabs/SensioGeneratorBundle/pull/468#issuecomment-181878810.

PR #518, opened 2016-10-25
Fix a forgotten string format

williamdclt:
This is a simple bugfix, avoid a raw "%s" appearing in the terminal when this RuntimeException is thrown.

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -350,7 +350,7 @@ EOT
             $manipulator->addResource($bundle);
         } catch (\RuntimeException $e) {
             return array(
-                '- Import the bundle\'s %s resource in the app\'s main configuration file:',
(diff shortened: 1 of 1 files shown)


xabbuh:
:+1:

Command/GenerateBundleCommand.php
```diff
@@ -350,7 +350,7 @@ protected function updateConfiguration(OutputInterface $output, Bundle $bundle)
             $manipulator->addResource($bundle);
         } catch (\RuntimeException $e) {
             return array(
-                '- Import the bundle\'s %s resource in the app\'s main configuration file:',
+                sprintf('- Import the bundle\'s %s resource in the app\'s main configuration file:', $bundle->getServicesConfigurationFilename()),
```
HeahDude: Thanks for submitting this fix :) Can you please surround `%s` with quotes? Then it would be perfect, thank you very much!

fabpot: I will do it myself.

fabpot:
Thanks for fixing this bug @williamdclt.

PR #526, opened 2016-11-08
Minor grammar corrections in bundle generator command

darrylhein:


Command/GenerateBundleCommand.php
```diff
@@ -150,7 +150,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
                 'See http://symfony.com/doc/current/cookbook/bundles/best_practices.html#bundle-name for more',
                 'details on bundle naming conventions.',
                 '',
-                'Use <comment>/</comment> instead of <comment>\\ </comment> for the namespace delimiter to avoid any problem.',
+                'Use <comment>/</comment> instead of <comment>\\</comment> for the namespace delimiter to avoid any problems.',
```
HeahDude: the space after the backslash should be reverted and the one before `for` removed instead, or `<` will be escaped breaking the style.

Command/GenerateBundleCommand.php
```diff
@@ -150,7 +150,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
                 'See http://symfony.com/doc/current/cookbook/bundles/best_practices.html#bundle-name for more',
                 'details on bundle naming conventions.',
                 '',
-                'Use <comment>/</comment> instead of <comment>\\ </comment> for the namespace delimiter to avoid any problem.',
+                'Use <comment>/</comment> instead of <comment>\\ </comment>for the namespace delimiter to avoid any problems.',
```
xabbuh: What about removing the space inside the `comment` tag instead?

HeahDude: Does not work see the outdated diff with my comment

xabbuh: Shouldn't four backslashes do the trick then?

HeahDude: If you ask me I'd keep like this, but I've got no hard feeling :)

fabpot:
Thank you @darrylhein.

Command/GenerateBundleCommand.php
guilliamxavier: For reference: #141, #137.

PR #527, opened 2016-11-19
Fixed Controller command not updating config routing file

### LINKED ISSUE #284: [DX] Potential missing annotation route import
I _believe_ there is a bug when using `doctrine:generate:crud` with the annotation format. Specifically, the routing import (to import the `Controller` directory with `type: annotation`) is not added. I believe you can see this here: https://github.com/sensiolabs/SensioGeneratorBundle/blob/master/Command/GenerateDoctrineCrudCommand.php#L122

My guess is that the thought is that if you originally generated the bundle with the annotation config format, then you'll already have this import. We should do a little more work to make sure that the controller _is_ in fact being imported in routing. Without it, the CRUD seems to not work at first.

Thanks!

HeahDude:
Closes #284.

### DIFF
--- a/Command/GenerateControllerCommand.php
+++ b/Command/GenerateControllerCommand.php
@@ -11,6 +11,7 @@
 
 namespace Sensio\Bundle\GeneratorBundle\Command;
 
+use Sensio\Bundle\GeneratorBundle\Manipulator\RoutingManipulator;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


Command/GenerateControllerCommand.php
```diff
@@ -333,4 +346,19 @@ protected function createGenerator()
     {
         return new ControllerGenerator($this->getContainer()->get('filesystem'));
     }
+
+    /**
+     * @param string $bundleName
+     * @param string $controller
+     */
+    protected function tryUpdateAnnotationRouting($bundleName, $controller)
```
fabpot: should be private, and phpdoc can be removed.

HeahDude: Done.

Command/GenerateControllerCommand.php
```diff
@@ -333,4 +346,15 @@ protected function createGenerator()
     {
         return new ControllerGenerator($this->getContainer()->get('filesystem'));
     }
+
+    private function tryUpdateAnnotationRouting($bundleName, $controller)
```
adrienlucas: Maybe this should go in `Sensio\Bundle\GeneratorBundle\Generator\ControllerGenerator::generateRouting()` ?

HeahDude: It seems that generators do not use manipulators, only commands uses both.

HeahDude:
ping @fabpot

Command/GenerateControllerCommand.php
adrienlucas: Fair enough ! (sorry for the latency)

fabpot:
Thank you @HeahDude.

PR #531, opened 2017-01-14
Fix typo Compoennt -> Component

juliendufresne:


fabpot:
Good catch, thanks @juliendufresne.

PR #533, opened 2017-01-19
Update generate_bundle.rst

skydiablo:


HeahDude:
Hello @skydiablo as you can see at the beginning of the file, we explain how to use the console, because this bundle is still compatible with symfony 2.x we'll keep it as is until next major release.

Hence I'm closing here, thank you for understanding.

PR #537, opened 2017-02-15
Fix command line snippets

djbb7:
The console script is located in the bin folder, not in the app folder.


HeahDude:
Hello @djbb7, thank you for your contribution. However, we already rejected proposition like this before, the reason is that there is a caution at the beginning of the file you've changed and the same goes for every other commands because Symfony 2.x and 3.x are both supported by the current version of this bundle.

Hence I'm closing here, thank you for understanding.

PR #540, opened 2017-03-02
Removed excesive space at the end of the line.

Briones:
When a Form is generated it shows a large space between the $builder->add line and the `;` at the end.

The code generated is:
```
    public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder->add('username')->add('password')        ;
    }
```

With the fix: 
```
    public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder->add('username')->add('password');
    }
```


fabpot:
Good catch, thanks @Briones.

PR #543, opened 2017-03-03
Symfony 3 console location updated and command logic improved

JasonBenett:
This is the SF3 documentation so it seems to be logical to use the bin/console folder by default. Moreover, the "generate:doctrine" format is not the one given by the "bin/console list" command so I reversed with "doctrine:generate".


HeahDude:
Thanks for your contribution, we already refuse this change many times in the past, and it's the same for all commands, but this bundle is still compatible with symfony 2.7 and 2.8 LTS versions.

About the command name, we'd like to deprecate commands not starting by `generate` in a next versions.

Hence I'm closing, thank you for understanding.

Issue #548, opened 2017-03-30
crud error with name entity singularized


gabrielrincon:
Hi everyone. When i used the console for generate a CRUD, in the controller have a bug with the name of the entity, for example if i used Examples entity, in the method new the first line is make:

$example = new Example();

the rigth sentences is

$example = new Examples();

Thanks very much

PR #550, opened 2017-04-03
Correction on plural name entitiy

gabrielrincon:
Because is a Bug with the entity name when i used a plural.  

Because, dont make the name class correctly. 

BUG: $tipoPersona = new Tipopersona();

CORRECT: $tipoPersona = new TipoPersona();

https://github.com/sensiolabs/SensioGeneratorBundle/issues/548

### DIFF
--- a/Resources/skeleton/crud/actions/new.php.twig
+++ b/Resources/skeleton/crud/actions/new.php.twig
@@ -16,7 +16,7 @@
 {% endblock method_definition %}
     {
 {% block method_body %}
-        ${{ entity_singularized }} = new {{ entity_singularized|capitalize }}();
+        ${{ entity_singularized }} = new {{ entity }}();
         {% if use_form_type_instance -%}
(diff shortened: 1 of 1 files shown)


gabrielrincon:
Today, persist this error. Please help me to approve this item. Thanks. @javiereguiluz @fabpot

gabrielrincon:
Look this error. My entity is called Pais. An this bug make me this change.

captura de pantalla de 2017-07-11 11-14-49

OskarStark:
You should rename your entity to a name which could be pluralized

fabpot:
@OskarStark is right. Closing.

javiereguiluz:
Quick comment: this has caused a lot of issues in the past. It's easy to think that people always develop applications in English ... but that's not true. We've received similar reports from people developing in French, Spanish, etc.

My question: could we consider changing this "feature" for Symfony 4? Thanks!

gabrielrincon:
@javiereguiluz thanks very much.

PR #552, opened 2017-05-09
change to bin/console

diaboloshogunate:
symfony 3 uses `bin/console` instead of `app/console`. It is stated in the caution box but since this doc is for symfony 3.x we should show the `bin/console` in the examples instead of the old way.


fabpot:
Thank you @diaboloshogunate.

PR #553, opened 2017-05-18
Allow Symfony 4

dunglas:


fabpot:
Not sure about this one as the structure in Symfony 4 (via Flex) will be very different, so I doubt that this will work just by adding this.

dunglas:
@fabpot still useful if you use the standard edition with Symfony 4 (legacy projects for instance).

cbdiesse:
I don't understand why Symfony 4 applications cannot use bundles for modularity and features extension for large web applications. flex not seem concerned by those features and then internal bundle generator must not be affected unless S4 bundles directory structure is actually unknown.

fabpot:
Closing as symfony/maker-bundle should be used instead. Standard edition is not available for Symfony 4 anyway.

PR #557, opened 2017-07-10
Explain recent composer.json changes

jezemery:
Recent changes to the composer file in the Symfony Framework Standard editon cause (bin/app)/console generate:bundle to fail with an "unclear" error message.

Change is here https://github.com/symfony/symfony-standard/commit/5f7ef8c34e2c947ccb447da24bc287beccacb443

This addition will explain the message to those looking for a solution

### DIFF
--- a/Resources/doc/commands/generate_bundle.rst
+++ b/Resources/doc/commands/generate_bundle.rst
@@ -28,6 +28,21 @@ forget to pass all needed options:
 
     $ php app/console generate:bundle --namespace=Acme/Bundle/BlogBundle --no-interaction
 
+.. caution::
+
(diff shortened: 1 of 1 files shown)


jezemery:
This needs to be updated for the following versions
```v3.3.4 v3.3.3 v3.2.11 v3.2.10 v2.8.24 v2.8.23 v2.7.31 v2.7.30 ```

**Comment by OskarStark** [comment]
Travis is failing due to a deprecation warning:

````
The Symfony\Component\Process\ProcessBuilder class is deprecated since version 3.4 and will be removed in 4.0. Use the Process class instead: 1x
    1x in KernelManipulatorTest::testAddToArray from Sensio\Bundle\GeneratorBundle\Tests\Manipulator
```

Pierstoval:
I think this warning should also be present if any generated class is not autoloadable after generation

fabpot:
Thank you @jezemery.

fabpot:
reworded in de22a60

PR #559, opened 2017-07-12
Make generate:bundle report errors again

### LINKED ISSUE #558: generate:bundle doesn't report errors (regression)
On a new symfony-standard project, with `composer.json` containing the following lines for autoloading:

```json
    "autoload": {
        "psr-4": {
            "AppBundle\\": "src/AppBundle"
        },
        "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
    },
```

let's generate a new bundle with the command line

```
php bin/console generate:bundle --namespace=BlogBundle --no-interaction
```

on two distinct installs:

1. with vendor sensio/generator-bundle before commit ddf8b6d1f857b17562b1e6e34a81c86a5eb79dd1, e.g. v3.0.11:

397-a

i.e. it reports an error _which is expected (see #555)_.

2. with vendor sensio/generator-bundle after commit ddf8b6d1f857b17562b1e6e34a81c86a5eb79dd1, e.g. 3.1.0 (or current v3.1.4):

397-b

i.e. it **no longer reports any error**, but any subsequent attempt to boot the app kernel **will still fail**:

397-b2

so that's probably an unintended side effect of PR #397.

---

Note: issue #555 is about the generated bundle not being autoloaded automatically, whereas the present issue is about the command not reporting the error (while it did before).

guilliamxavier:
Simply revert a regression in #397 after which errors returned by sub-calls were ignored.
fixes #558

### DIFF
--- a/Command/GenerateBundleCommand.php
+++ b/Command/GenerateBundleCommand.php
@@ -91,19 +91,20 @@ EOT
         $generator->generateBundle($bundle);
 
         $errors = array();
+        $runner = $questionHelper->getRunner($output, $errors);
 
         // check that the namespace is already autoloaded
-        $this->checkAutoloader($output, $bundle);
+        $runner($this->checkAutoloader($output, $bundle));
 
         // register the bundle in the Kernel class
-        $this->updateKernel($output, $this->getContainer()->get('kernel'), $bundle);
+        $runner($this->updateKernel($output, $this->getContainer()->get('kernel'), $bundle));
 
         // routing importing
-        $this->updateRouting($output, $bundle);
+        $runner($this->updateRouting($output, $bundle));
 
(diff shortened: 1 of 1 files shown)


guilliamxavier:
New output:

397-c

---

Please note that I did not introduce any other change, so, in particular:

- the `OK`/`FAILED` lines are not indented
- there is no `OK` line after the "`Generating a sample bundle skeleton into` ..." section (see https://github.com/sensiolabs/SensioGeneratorBundle/commit/ad28da85e9195ba73e5dd4811bc7f439b89da6ee#diff-fc3746671a611dd19b80d15471463f48L115, and `BundleGenerator::generateBundle()` doesn't return errors but throws exceptions)
- there is a trailing colon at the end of the message (see commit 3d16db1631a3648d1b6662efe6de32735118b1da)

PR #391, opened 2015-07-17
replaced hardcoded command names in help

OskarStark:
and moved command names and description to the top of the configure()-method

PR #560, opened 2017-07-13
Replace hardcoded command name and "app/console"

guilliamxavier:
This file did not exist yet at the time of #391.


fabpot:
Thank you @guilliamxavier.

PR #562, opened 2017-07-13
Fix formatting + mention `composer dump-autoload`

guilliamxavier:
This is a follow-up to #557 (and commit de22a60ab53cd6edc62fdb121788fd469afff224).

### DIFF
--- a/Resources/doc/commands/generate_bundle.rst
+++ b/Resources/doc/commands/generate_bundle.rst
@@ -32,17 +32,26 @@ forget to pass all needed options:
 
     If the ``generate:bundle`` command returns an error about registering the
     bundle namespace in ``composer.json``, add the following line to your
-    ``composer.json`` file within the ``psr-4`` section
-    ``"Acme\\Bundle\\BlogBundle\\": "src/Acme/Bundle/BlogBundle"``:
+    ``composer.json`` file within the ``psr-4`` section:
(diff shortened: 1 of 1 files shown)


Resources/doc/commands/generate_bundle.rst
```diff
@@ -32,17 +32,26 @@ forget to pass all needed options:
...
+            "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
         },
-        "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
-    },
+
+    ), then execute the following command to regenerate the autoload files:
+
+    .. code-block:: bash
+
+        $ composer dump-autoload
```
fabpot: This should not be needed in dev, so I would remove this part.

guilliamxavier: @fabpot I don't know if it should, but it seems needed even in dev (that or a `composer install`, which includes autoloader dump), see for example https://pastebin.com/CF0DEQbt and https://pastebin.com/hy04dQpX
(by the way, in prod we would probably add `-o`)

fabpot: hmmm, you're right

fabpot:
Thank you @guilliamxavier.

PR #566, opened 2017-07-22
Fix small typo

mercuryseries:
Use equality symbol instead of colon.


fabpot:
Thank you @mercuryseries.

Issue #544, opened 2017-03-04
CRUD Controller Imports in one line.


stepotronic:
When creating CRUD with write actions the use-statement for Request is added directly after the Route import:
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;use Symfony\Component\HttpFoundation\Request;
I believe it is because of the tokenizing of the two following conditions in the original template.  In my fork i added a {{ "\n" }} in the beginning of line 16 in the "Resources/skeleton/crud/controller.php.twig"

fabpot:
Closing this issue as this bundle is now obsolete. Please, use the maker bundle instead.

PR #572, opened 2017-08-11
Fix Crud Controller imports in one line

matthieumota:
As explained in #544, there is a little bug when generate a CRUD with annotations and write actions who need Request class import.


Resources/skeleton/crud/controller.php.twig
```diff
@@ -13,6 +13,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 {%- endif %}
 {% if 'new' in actions or 'edit' in actions or 'delete' in actions %}
+{{ ('annotation' == format) ? "\n" -}}
```
HeahDude: What about removing the white space control line 11 instead?

matthieumota:
@fabpot I close that because I suppose bundle will be unmaintained (With Maker and 2.8 EOL at end of year)

PR #584, opened 2017-12-06
Updated displaying of the "Summary before generation" in some commands

AndrewBu:
I updated displaying of the "Summary before generation" section in the GenerateControllerCommand and GenerateDoctrineCrudCommand. There was a small typo ("bg=blue;fg-white") in the GenerateControllerCommand class. It was the cause of different displaying of the "Summary before generation" section in comparison with other sections like "Welcome to the Symfony controller generator" or "Controller generation". Especially the "Summary before generation" section was unreadable in PhpStorm. 
It looked this way:
incorrect_section_presentation
There is how it looked in PhpStorm in the bottom of the screenshot and how it looked in the Linux console in the top of the screenshot.
The GenerateDoctrineCrudCommand class did not have the same error but i updated the "Summary before generation" section there to make the section in these classes the same.

### DIFF
--- a/Command/GenerateCommandCommand.php
+++ b/Command/GenerateCommandCommand.php
@@ -128,10 +128,8 @@ EOT
         }
 
         // summary and confirmation
+        $questionHelper->writeSection($output, 'Summary before generation');
         $output->writeln(array(
-            '',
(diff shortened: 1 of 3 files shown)


fabpot:
Thank you @AndrewBu.

PR #593, opened 2018-04-07
Added autocomplete for routes

devtronic:
Added auto completion for the routes based on previous entered routes.

This is helpful if you create a lot of similar routes like
- `/admin/user`
- `/admin/user/{id}`
- `/admin/user/{id}/edit`
- ...


fabpot:

PR #595, opened 2018-07-18
[MOD] update skeletion crud actions new.php.twig

lcp0578:
When entity name like InfoCategory, 
```
public function newAction() {
       $infoCategory = new Infocategory(); //Class is not found.
      $infoCategory = new InfoCategory(); //not c is C
}
```


fabpot:

PR #596, opened 2018-11-13
@Method annotation removed in Controller's actions in order to have new "methods" argument inside @Route

MiguelAlcaino:
Hi.
I found out that the `doctrine:generate:crud` task was using the old `@Method` annotation for generated actions inside Controllers. I updated the `use` inside the controllers skeletons and also the `@Method` annotation was removed in order to use the new argument `methods` inside the annotation `@Route` .

Because of this, functionalities like the delete button for a database record was not working due to the use of the deprecated `@Method` annotation. 

I also fixed a `datetime` assertion that was making one of the tests to fail. And updated the `testGenerateAnnotation` in DoctrineCrudGeneratorTest.php to verify that the new generateds annotations are working find and to verify that the old ones are not used anymore.

In travis the build is failing due to deprecations.
selection_059


fabpot:

PR #597, opened 2018-11-21
Default to installing bundle as dev dependency

darrylhein:
Seems to me that's what people would want, but could be wrong.


fabpot:

